PR #8, opened 2019-03-26
TMPFIX: return empty edge set for out-of-range node IDs

arsenius7:


src/edges.cpp
```diff
@@ -63,6 +63,12 @@ Selection _resolveIndex(const HighFive::Group& indexGroup, const NodeID nodeID)
 {
     typedef std::vector<std::vector<uint64_t>> RawIndex;
 
+    if (nodeID >= indexGroup.getDataSet("node_id_to_ranges").getSpace().getDimensions()[0]) {
+        // Returning empty set for out-of-range node IDs, to be aligned with SYN2 reader implementation
```
mgeplf: I'm not sure our syn2 implementation should be bleeding into the sonata impl.  Can we get the syn2 format to throw an exception instead?

arsenius7: I'm afraid that might break simulation part, if all SYN2 files are used for every GID.

mgeplf: Ok, I guess the semantics make sense.  I don't think this will be a temporary fix though, once this behavior is considered the norm.  I'm ok with that, though

arsenius7: Come on, be an optimist. :)
Should we ever make our mind about node population IDs (whether they should be implicit, use some offset or explicit mapping), we will inevitably revisit this part.

PR #14, opened 2019-04-26
Travis build matrix with Ubuntu and Mac OS

wizmer:


.travis.yml
```diff
@@ -20,4 +35,5 @@ install:
   - pip install nose .
 
 script:
-  - nosetests python
+  - nosetests python &&
+    rm -rf build && mkdir build && cd build && cmake -DCMAKE_BUILD_TYPE=Release -DEXTLIB_FROM_SUBMODULES=ON .. && make && make test
```
arsenius7: Is there any difference between
```
script:
  - A
  - B
```
and
```
script:
  - A && B
```
?

wizmer: From what I was able to see, the build is only reported as failure if the last line has a non zero error code.

PR #15, opened 2019-04-26
Get version from a VERSION file

wizmer:
In order to be able to pass the version as an argument of setup().
Otherwise it is not possible because, cmake needs to be run first. And this happens only inside the call of setup()


arsenius7:
I would rather prefer CMake-generated `__init__.py` with version constant (I don't think we need version in C++ so far, but don't mind keeping it).

PR #19, opened 2019-05-01
Add flag -Dsonata_CXX_WARNINGS to disable Werror in setup.py

wizmer:


setup.py
```diff
@@ -74,6 +74,7 @@ def build_extension(self, ext):
             "-DEXTLIB_FROM_SUBMODULES=ON",
             "-DSONATA_PYTHON=ON",
             "-DCMAKE_BUILD_TYPE=",
+            "-Dsonata_CXX_WARNINGS=OFF",
```
mgeplf: capitalize

CMakeLists.txt
```diff
@@ -12,6 +12,7 @@ project(sonata VERSION ${SONATA_VERSION})
 
 option(EXTLIB_FROM_SUBMODULES "Use Git submodules for header-only dependencies" OFF)
 option(SONATA_PYTHON "Build Python extensions" OFF)
+option(${PROJECT_NAME}_CXX_WARNINGS "Compile C++ with warnings" ON)
```
mgeplf: do you want to have the option be turn off warnings, or 'warnings are errrors'?
I don't think it's bad to spew warnings, it's just that you don't want the build to fail when building w/ python, right?

PR #22, opened 2019-05-21
Fixed MANIFEST.in

arsenius7:
* include VERSION file
 * include third-party libraries

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -26,13 +26,14 @@ cache:
     - $HOME/Library/Caches/Homebrew
 
 before_install:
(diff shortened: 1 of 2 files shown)


arsenius7:
Is there a way we can check installation from source tarball in CI?

wizmer:
Here https://github.com/BlueBrain/libsonata/blob/6d7e3929c156812969d9237bd797bf6aba6e50ec/.travis.yml#L35
you could `python setup sdist` first and intstall from here instead of `pip install .`

PR #23, opened 2019-05-21
Fixed direct Python install (`make install`)

arsenius7:
* to revive 'libsonata+python' Spack package

### DIFF
--- a/python/CMakeLists.txt
+++ b/python/CMakeLists.txt
@@ -1,5 +1,3 @@
-set(SONATA_PYTHON_BUILD ${CMAKE_CURRENT_BINARY_DIR}/libsonata)
-
 if (EXTLIB_FROM_SUBMODULES)
     add_subdirectory(pybind11 EXCLUDE_FROM_ALL)
 else()
@@ -18,3 +16,10 @@ target_link_libraries(sonata_python
     PRIVATE fmt::fmt-header-only
     PRIVATE pybind11::module
 )
+
+if (PYTHON_INSTALL_SUFFIX)
+    install(TARGETS sonata_python
+        LIBRARY
+        DESTINATION ${PYTHON_INSTALL_SUFFIX}
+    )
+endif()


arsenius7:
Alas, looks like that's not enough.
For some reason, the entry points for any package depending on `libsonata` can not recognize `libsonata.cpython-36m-x86_64-linux-gnu.so` file.
I.e., assuming this file can be found in folder `foo`, this works:
```
PYTHONPATH=foo python -c "import libsonata"
```
but this:
```
PYTHONPATH=foo brainbuilder --help
```
ends up with
```
pkg_resources.DistributionNotFound: The 'libsonata<1.0,>=0.0.1' distribution was not found and is required by voxcell
```

`libsonata+python` and its dependencies thus remain broken in Spack.

arsenius7:
Also, in Spack world
```
python -c "import libsonata"
```
actually gives me
```
ImportError: libhdf5.so.103: cannot open shared object file: No such file or directory
```
unless I explicitly add `hdf5` as a dependency to `libsonata` recipe.
Can't say if it was the case before or not.

wizmer:
To be honest, I've never been able to correctly install the python package with the old recipe. And that's the first reason that made me create the `pip install` recipe.

I always have (with this commit or with old commits before I changed the recipe):
```
$ python -c "import libsonata"
Traceback (most recent call last):
  File "<string>", line 1, in <module>
ImportError: dynamic module does not define module export function (PyInit_libsonata)
```

arsenius7:
> ImportError: dynamic module does not define module export function

That started to happen when Python module and C++ library name were made same.
When `python -c "import libsonata"` is ran from the folder where C++ `libsonata.so` is located, that one is picked up.

CMake flags example are available in the Spack recipe; where `PYTHON_INSTALL_SUFFIX` is the path to `site-packages` folder.

arsenius7:
The key point is that broken `libsonata+python` in Spack world is now a blocker for circuit build pipeline updates; and I would rather not fix it by pinning some old `libsonata` commit there.

Given that `make install` option is not working for Python as it used before (at https://github.com/BlueBrain/spack/pull/350 we believed it does), another try with `py-libsonata` makes sense for me, as there is already `py-morphio` and `py-basalt`.

arsenius7:
https://github.com/BlueBrain/spack/pull/372 should revive `libsonata` Python package in Spack; and I'll remove the leftovers of `make install` in another PR.

Issue #2, opened 2019-03-13
Edge indexing tool


arsenius7:
We need a reusable component for [edge indexing](https://github.com/AllenInstitute/sonata/blob/master/docs/SONATA_DEVELOPER_GUIDE.md#edges---optional-indexing).

Ideally, it should be available in two flavors: as a method in API, and as a console command.

PR #28, opened 2019-05-28
Naive edge indexing support [#2]

arsenius7:
Since it is a "naive" implementation, a number of shortcuts were taken:
- reading source datasets fully into memory, no chunking
- keeping indices in memory
- no parallelization
- no progressbar
- no cleanup on failure (at quick glance I have not found "remove" method in HighFive)

With regard to parallelization, I messed around with MPI-based parallelized indexing in `syn-tool`, but on a "medium" file of 10^9 synapses (~60G with properties), have not seen any speedup (on contrary, on 4 nodes I got x4 _slowdown_). It's not enough to make any conclusion but still enigmatic.


tests/test_edges.cpp
```diff
@@ -116,6 +118,61 @@ TEST_CASE("EdgePopulation", "[edges]")
...
+    dst << src.rdbuf();
+}
+
+}  // unnamed namespace
+
+
+TEST_CASE("EdgePopulation::writeIndices", "[edges]")
+{
+    const std::string srcFilePath = "./data/edges-no-index.h5";
+    const std::string dstFilePath = "./data/edges-no-index.h5.tmp";
```
arsenius7: another shortcut: no "proper" temporary file (once again, postponed until `std::filesystem` is available)

matz-e:
> With regard to parallelization, I messed around with MPI-based parallelized indexing in `syn-tool`, but on a "medium" file of 10^9 synapses (~60G with properties), have not seen any speedup (on contrary, on 4 nodes I got x4 _slowdown_). It's not enough to make any conclusion but still enigmatic.

I'm wondering about the motivation to not use any MPI here… (I understand that no MPI may simplify some stuff)

Do we care about absolute raw speed for tiny circuits, or scalability on the larger ones? What are the memory limits here in terms of neurons in the circuit?

arsenius7:
> I'm wondering about the motivation to not use any MPI here… (I understand that no MPI may simplify some stuff)

Yes, the main motivation here is to keep the initial implementation as simple as possible, and merely to provide at least _something_ for indexing edges in `libsonata` (as mentioned above, serial implementation is not the only shortcut taken).

As of now, even such a naive implementation covers most of our current usecases (for instance, no edge file for `Isocortex/20190307` exceeds 4 x 10^9 edges).

> Do we care about absolute raw speed for tiny circuits, or scalability on the larger ones? What are the memory limits here in terms of neurons in the circuit?

You are right, of course, raw speed for tiny circuits is not the goal, and eventually we need the thing to scale for larger ones; HPC team expertise here would be indispensable (no jokes).

I merely skimmed through MPI-based indexing implementation in `syntool`, but I have a feeling that only initial scan through node ID values is done there in parallel, and then all the resulting ranges are gathered in memory on one machine, so it does not provide out-of-core guarantees either. Nor is it obvious for me how to avoid this "reduce" step.

On the other hand, this indexing task can be easily described with MapReduce paradigm:

1) scan through nodeIDs, generate (nodeID, position) pairs
2) group pairs by nodeID, sort and aggregate ranges for each group
3) gather (nodeID, ranges), sort by nodeID, split into primary and secondary index levels

so maybe we can index with Spark? (I bet it can be done more efficiently than 1)-3) ).

We also get sorting by source / target node IDs prior to indexing "for free" there (Spark Functionalizer already does that, right?).

Finally, as a sidenote, I have some doubts about scalability of the format _itself_ the way it is now, no matter how smart the implementation is; at least for bidirectional access. So maybe that's where the effort would go first when it comes to huge synapse collections. :)

src/edge_index.cpp
```diff
@@ -106,6 +111,127 @@ Selection resolve(const HighFive::Group& indexGroup, const std::vector<NodeID>&
...
+std::unordered_map<NodeID, RawIndex> _groupNodeRanges(const std::vector<NodeID>& nodeIDs)
+{
+    std::unordered_map<NodeID, RawIndex> result;
+
+    if (nodeIDs.empty()) {
+        return result;
+    }
+
+    uint64_t i0 = 0;
+    NodeID lastNodeID = nodeIDs[0];
```
mgeplf: nit: `nodeIDs[id0]`, to show what i0 is used for.  I'd consider renaming `i0` `group_start_idx` or something, personally

src/edge_index.cpp
```diff
@@ -106,6 +111,127 @@ Selection resolve(const HighFive::Group& indexGroup, const std::vector<NodeID>&
...
+
+
+void _writeIndexGroup(const std::vector<NodeID>& nodeIDs, NodeID maxNodeID, HighFive::Group& h5Root, const std::string& name)
+{
+    auto indexGroup = h5Root.createGroup(name);
+
+    auto nodeToRanges = _groupNodeRanges(nodeIDs);
+    const auto rangeCount = std::accumulate(
+        nodeToRanges.begin(), nodeToRanges.end(), uint64_t(0),
+        [](uint64_t total, const std::pair<NodeID, RawIndex>& item) {
```
mgeplf: `const std::pair<NodeID, RawIndex>&` -> `decltype(nodeToRanges)::const_reference item` ?

src/edge_index.cpp
```diff
@@ -106,6 +111,127 @@ Selection resolve(const HighFive::Group& indexGroup, const std::vector<NodeID>&
...
+    uint64_t offset = 0;
+    for (NodeID nodeID = 0; nodeID < maxNodeID; ++nodeID) {
+        const auto it = nodeToRanges.find(nodeID);
+        if (it == nodeToRanges.end()) {
+            primaryIndex.push_back({offset, offset});
+        } else {
+            auto& ranges = it->second;
+            primaryIndex.push_back({offset, offset + ranges.size()});
+            offset += ranges.size();
+            std::move(
```
mgeplf: nice, I think hardly anyone uses this overload sometimes

tests/test_edges.cpp
mgeplf: can you add the same TODO wrt c++17 about std::filesystem?

src/edge_index.cpp
arsenius7: OK

src/edge_index.cpp
arsenius7: Thanks for suggesting a C++11 alternative.
I guess it would appeal more to people following modern C++; but might confuse those less savvy (I would apply it nevertheless). ^^

arsenius7:
@matz-e , I will merge this PR as it is for now, to unblock some work dependent on it; but we would heed to any after-merge comments you have.
As already stated, when / if it comes to optimizing `libsonata`, we mostly put our hopes into HPC team. ^^

src/edge_index.cpp
```diff
@@ -168,11 +168,11 @@ void _writeIndexGroup(const std::vector<NodeID>& nodeIDs, NodeID maxNodeID, High
     RawIndex primaryIndex;
     RawIndex secondaryIndex;
 
-    primaryIndex.reserve(maxNodeID);
```
arsenius7: `maxNodeID` was actually used as node count

PR #29, opened 2019-06-04
Optimized some C++/Python transitions

arsenius7:


include/bbp/sonata/population.h
```diff
@@ -48,6 +50,33 @@ class SONATA_API Selection
     const Ranges ranges_;
 };
 
+template<typename Iterator>
```
arsenius7: moved from `population.cpp`

python/bindings.cpp
```diff
@@ -299,7 +299,10 @@ PYBIND11_MODULE(libsonata, m)
             "Selection from list of intervals"
         )
         .def(
-            py::init(&Selection::fromValues),
+            py::init([](py::array_t<uint64_t, py::array::c_style | py::array::forcecast> values) {
```
arsenius7: I have not found a "canonical" way to get (begin, end) iterator pair from NumPy array object in `pybind11`.
I _hope_ that casting to C-style array and using raw pointers as iterator pair is safe.
Alternatively, we can make an iterator class ourselves but that seems to be a bit of an overkill.

arsenius7: 20s -> 0.4s for 10^8 ids

python/bindings.cpp
```diff
@@ -310,8 +313,10 @@ PYBIND11_MODULE(libsonata, m)
         )
         .def(
             "flatten",
-            &Selection::flatten,
-            "List of IDs constituting Selection"
+            [](Selection& obj) {
```
arsenius7: 3s -> 0.34s for 10^8 ids.
Replacing `reserve(N)` + `push_back` with `std::vector X(N)` + `std::iota` in `flatten()` C++ implementation reduces it further to 0.26s, but I guess these fractions of second are irrelevant.

python/test.py
```diff
@@ -3,17 +3,19 @@
...
+
 class TestSelection(unittest.TestCase):
     def test_basic(self):
         ranges = [(3, 5), (0, 3)]
         selection = Selection(ranges)
         self.assertTrue(selection)
         self.assertEqual(selection.ranges, ranges)
         self.assertEqual(selection.flat_size, 5)
-        self.assertEqual(selection.flatten(), [3, 4, 0, 1, 2])
+        self.assertEqual(selection.flatten().tolist(), [3, 4, 0, 1, 2])
```
wizmer: assert_array_equal instead ?

arsenius7: You mean, `numpy.testing`, right?
`tolist()` in effect checks also the result type (i.e. that it is an `array`, not a `list`).
This time maybe it's a desired effect.

wizmer: Ok, I see

python/bindings.cpp
mgeplf: isn't there a chance that the NumPy array isn't contiguous: ie if it's a slice into another array, the iteration may need to have a strides setup.

I think it's safe to do what you're doing, as long as there's a way to check for contiguousness, and fall back to the slow path otherwise.

python/bindings.cpp
mgeplf: nice

python/bindings.cpp
mgeplf: very nice

arsenius7: > isn't there a chance that the NumPy array isn't contiguous

As far as I understand, that's what `py::array::forcecast` is aimed to ensure:
https://pybind11.readthedocs.io/en/stable/advanced/pycpp/numpy.html#arrays

I haven't tried with a strided array, guess it's worth checking.

mgeplf: Ah, thanks.  That sounds like it does the right thing, sorry about the noise.

arsenius7: On contrary, it's easy to get trapped by false assumptions with this stuff.
I have added some unit tests at #30.
Truth be told, I'm still not 100% sure it does the right thing. ^^

PR #30, opened 2019-06-05
Added some Python unit tests

arsenius7:
* checking strided NumPy array
   as Selection::fromValues() argument

### DIFF
--- a/python/test.py
+++ b/python/test.py
@@ -1,6 +1,8 @@
 import os
 import unittest
 
+import numpy as np
+
 from libsonata import *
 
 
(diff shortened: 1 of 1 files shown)


python/test.py
```diff
@@ -18,8 +20,15 @@ def test_basic(self):
...
-        self.assertEqual(selection.ranges, [(1, 2), (3, 5), (1, 2)])
+        values = [
+            [1, 3, 4, 1],
+            [0, 0, 0, 0],
+        ]
+        expected = [(1, 2), (3, 5), (1, 2)]
+        self.assertEqual(Selection(values[0]).ranges, expected)
+        self.assertEqual(Selection(np.array(values, dtype=np.uint64, order='C')[0]).ranges, expected)
+        self.assertEqual(Selection(np.array(values, dtype=np.uint32, order='C')[0]).ranges, expected)
+        self.assertEqual(Selection(np.array(values, dtype=np.uint64, order='F')[0]).ranges, expected)
```
arsenius7: fails without `py::array::c_style | py::array::forcecast` incantation

wizmer: Does it work if `values = []` ?

arsenius7: can't see why it shouldn't, but I have added a unit test just in case

PR #38, opened 2019-07-15
make data/generate.py python3 compatible

### LINKED ISSUE #37: Test data script is incompatible with Python 3
Python 2 will be EOL end of this year, and most of the community has moved to Python 3.

We should make sure that the data creating script works with 3 to be able to drop support for Python 2, too.

mgeplf:
fixes #37

### DIFF
--- a/tests/data/generate.py
+++ b/tests/data/generate.py
@@ -1,11 +1,21 @@
 #!/usr/bin/env python
 
+import sys
+
 import numpy as np
 import h5py
 
 
 def write_population_one_group(pop, prefix):
-    string_dtype = h5py.special_dtype(vlen=unicode)
+
+    if sys.version_info[0] == 2:
+        vlen_str_type = unicode
+    elif sys.version_info[0] == 3:
+        vlen_str_type = str
+    else:
+        raise 'Only python2/3 supported'
+
+    string_dtype = h5py.special_dtype(vlen=vlen_str_type)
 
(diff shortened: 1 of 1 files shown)


tests/data/generate.py
```diff
@@ -1,11 +1,21 @@
...
 
 def write_population_one_group(pop, prefix):
-    string_dtype = h5py.special_dtype(vlen=unicode)
+
+    if sys.version_info[0] == 2:
+        vlen_str_type = unicode
+    elif sys.version_info[0] == 3:
+        vlen_str_type = str
+    else:
+        raise 'Only python2/3 supported'
```
matz-e: You can avoid `sys` completely by just using:
```
try:
    vlen_str_type = unicode
except:
    vlen_str_type = str
```

mgeplf: True, but this is more clear why we're doing it, and means that if unicode ever shows up again, it's future proof

tomdele: I would agree with @mgeplf here. And broad exceptions are also bug prone.

matz-e: Fair enough! The exception could be specialized… anyways, this could be simplified again once Py2 has died.

mgeplf: yeah, that's the other reason I do it; easier to grep for `version_info` once we deprecate py2

tomdele: > The exception could be specialized…
Yes of course but you will have to use a NameError exception and I am always a bit reticent to use it in this particular use case (maybe I should not ...)

PR #39, opened 2019-07-15
feature: enable custom enumerations.

### LINKED ISSUE #36: Add support for enum datatypes
See the excellent specification document of [SONATA](https://github.com/AllenInstitute/sonata/blob/master/docs/SONATA_DEVELOPER_GUIDE.md#nodes---enum-datatypes) – from the HPC side, we need support for the enum datatypes for circuits, to avoid repeatedly calculating unique values and indices.

I propose to augment the population interface by adding/changing the following methods:
```
    /**
     * Get attribute values for given Selection
     *
     * Enum datatypes will be resolved.
     *
     * @param name is a string to allow attributes not defined in spec
     * @throw if there is no such attribute for the population
     * @throw if the attribute is not defined for _any_ element from the selection
     */
    template <typename T>
    std::vector<T> getAttribute(const std::string& name, const Selection& selection) const;

    /**
     * Get numerical enumation values for Attribute, given Selection
     *
     * @param name is a string to allow attributes not defined in spec
     * @throw if there is no such enum datatype attribute for the population
     * @throw if the attribute is not defined for _any_ element from the selection
     */
    std::vector<std::size_t> getAttributeIndices(const std::string& name, const Selection& selection) const;

    /**
     * Get list of the enumeration values for Attribute, given Selection
     *
     * @param name is a string to allow attributes not defined in spec
     * @throw if there is no such enum datatype attribute for the population
     * @throw if the attribute is not defined for _any_ element from the selection
     */
    std::vector<std::string> getAttributeValues(const std::string& name, const Selection& selection) const;
```

This is a blocker to use `libsonata` in circuit building.

matz-e:
@mgeplf after some consideration, I wasn't really happy with the scoping of the interface discussed in #36.

I've changed this to include:
```
enumerationNames
enumerationValues(name)
```
to have a clear distinction between attributes and these custom enumerations (these two are also top-level properties of a population). To access the resolved values, one can then use:
```
getAttribute<std::string>(name, selection)
```

Enumeration values can be explicitly requested via:
```
getEnumeration<integer_type>(name, selection)
getAttribute<integer_type>(name, selection)
```
The latter is more for Python support and to have a homogeneous interface between C++ and Python.

This way, circuits can be read in a simplistic way without relying explicitly on enumerations, and mapping will happen "magically".

Not sure if this is any better? The names certainly look easier to grok.

Fixes #36.


matz-e:
@mgeplf who can review this? HPC should get some more access rights if this is drifting our way.

mgeplf:
I can have a look, didn't know it was ready.  Sorry about the delay, if it was.

Totally agree about HPC having access.  I don't believe I have permission to do that, but I will hopefully find someone who does.

matz-e:
Thanks! I should have stated that this is ready.

src/population.cpp
```diff
@@ -176,6 +190,32 @@ std::vector<T> Population::getAttribute(const std::string& name, const Selection
...
 
+template<>
+std::vector<std::string> Population::getAttribute(const std::string& name, const Selection& selection) const
+{
+    if (impl_->attributeEnumNames.count(name) == 0) {
+        HDF5_LOCK_GUARD
+        return _readSelection<std::string>(impl_->getAttributeDataSet(name), selection);
+    }
+
+    auto indices = getAttribute<size_t>(name, selection);
```
mgeplf:
```suggestion
    const auto& indices = getAttribute<size_t>(name, selection);
```

same with `values`

src/population.cpp
```diff
@@ -176,6 +190,32 @@ std::vector<T> Population::getAttribute(const std::string& name, const Selection
...
+    auto values = enumerationValues(name);
+
+    std::vector<std::string> resolved;
+    resolved.reserve(indices.size());
+
+    auto max = values.size();
```
mgeplf: const

src/population.cpp
```diff
@@ -184,8 +224,27 @@ std::vector<T> Population::getAttribute(const std::string& name, const Selection
...
+
+    HDF5_LOCK_GUARD
+    return _readSelection<T>(impl_->getAttributeDataSet(name), selection);
+}
+
+
+std::string Population::_attributeDataType(const std::string& name, bool remap) const
 {
+    if (remap && impl_->attributeEnumNames.count(name) > 0) {
+        return "string";
```
mgeplf: Can we change `remap` to something more descriptive, please?  Maybe something inline with the doc in the .h - `translate_enum_type`?

src/population.cpp
```diff
@@ -244,6 +305,10 @@ INSTANTIATE_TEMPLATE_METHODS(uint32_t)
 INSTANTIATE_TEMPLATE_METHODS(int64_t)
 INSTANTIATE_TEMPLATE_METHODS(uint64_t)
 
+#ifdef __APPLE__
```
mgeplf: what happens w/o this?  Also, not knowing anything about macOS, is __APPLE__ the best way to catch this?

src/population.hpp
```diff
@@ -38,6 +39,19 @@ std::set<std::string> _listChildren(const HighFive::Group& group, const std::set
     return result;
 }
 
+std::set<std::string> _listCustomEnumerations(const HighFive::Group h5Group, const std::set<std::string>& attrs) {
+    std::set<std::string> names{};
+    std::set<std::string> enums(_listChildren(h5Group));
```
mgeplf: const

src/population.cpp
matz-e: Apparently, `__APPLE__` is a good way to detect Mac OS. It's mentioned [here](https://developer.apple.com/library/archive/documentation/Porting/Conceptual/PortingUnix/compiling/compiling.html) with some notes not to redefine functions.

Dropping it will lead to `size_t` duplicating `uint64_t` errors on other platforms. Do you have a better solution in mind?

mgeplf: I'm fine w/ the `__APPLE__` ifdef, just wanted to make sure it's the 'right' way.

matz-e: I think we could get around using macros by defining a "hidden function(s)" doing the instantiation. Then we could template this `#ifdef` out, and have a generic type-test if `size_t` and `uint64_t` are really different.

I may also be in favor of reducing macro usage.

mgeplf: No need to do it for this PR.  I'm ok w/ limited macro usage, and am comfortable w/ it in this case - they're clear, localized, and non-fancy.

src/population.hpp
```diff
@@ -38,6 +39,19 @@ std::set<std::string> _listChildren(const HighFive::Group& group, const std::set
     return result;
 }
 
+std::set<std::string> _listCustomEnumerations(const HighFive::Group h5Group, const std::set<std::string>& attrs) {
```
tomdele: Is the _list name on purpose even if the function returns a set ?

matz-e: List as in "to list", not referring to the actual datatype. Same goes for `_listChildren`, I guess.

tomdele: Just wanted to check with you.

include/bbp/sonata/population.h
```diff
@@ -97,9 +97,17 @@ class SONATA_API Population
      */
     const std::set<std::string>& attributeNames() const;
 
+    /**
+     * All attribute names that are custom enumerations
+     */
```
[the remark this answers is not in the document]
matz-e: @mgeplf better now?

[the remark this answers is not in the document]
mgeplf: I was hoping for something that would be C-f'able in the spec - to make it obvious that one can read up on theses 'enum' things there.  Either that, or having a link to the paragraph in the spec would be helpful for people reading the interface.

mgeplf:
> @mgeplf If your happy, I think this would be ready to merge and we should tag something like `0.1` for a release (IMO)

WRT release, I agree.  Currently the VERSION is at `0.0.3`.  I'm not sure we're at `0.1.0` yet, but we haven't expressed our versioning scheme either.  I also will have to look into how automated the release process is.  +1 for tagging it, though.

include/bbp/sonata/population.h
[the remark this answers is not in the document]
matz-e: I added links.

matz-e:
> I'm not sure we're at `0.1.0` yet, but we haven't expressed our versioning scheme either.

What are we missing?

mgeplf:
> What are we missing?

Multi-group populations, for instance.  Same with CSV support.  However, we don't have a roadmap for their support, so I guess we can just increment at will.

matz-e:
> I'm happy, do you want someone from HPC to review?

Added @tristan0x, but I would like to get this merged rather soon to proceed with updates to Spack, MVDTool, Spack, TouchDetector, Spykfunc (that's a nice pipeline I've got stacked up).

> Multi-group populations, for instance. Same with CSV support. However, we don't have a roadmap for their support, so I guess we can just increment at will.

I'm all for tagging `0.1` now, we are already relying on this. We should determine our needs for these features and then increment the minor as we add them. I guess full spec-implementation would warrant a `1.0`?

matz-e:
Proceeding to get this out and dependent packages updated.

PR #42, opened 2019-07-23
releases: switch to grabbing the version from tags.

### LINKED ISSUE #41: Remove VERSION file, use tags
We currently require `setuptools_scm`:

https://github.com/BlueBrain/libsonata/blob/98aa883849c700c369b532a9f4b6bf2a66ea4edc/setup.py#L146

With this¸ we don't need the dependency on the VERSION file any longer, and can just rely on the VCS to provide all necessary info . `setuptools_scm` will add information to Python tarballs to work without a `.git` directory, too.

For the C++ library, we can rely on `python setup.py --version` to get the version. See also here: https://github.com/BlueBrain/MVDTool/blob/master/CMakeLists.txt#L3

Supersedes #34 in a way.

matz-e:
Utilize already present dependency on `setuptools-scm` to extract
versions from git.

Closes #41.


setup.py
```diff
@@ -124,15 +124,11 @@ def run(self):
...
 
-with open('VERSION') as versionf:
-    version = versionf.readline().strip()
-
 setup(
     name="libsonata",
     description='SONATA files reader',
     author="BlueBrain Project, EPFL",
     author_email="bbp-ou-nse@groupes.epfl.ch",
-    version=version,
```
mgeplf: is `use_scm_version=True,` the default now?

matz-e: Right, forgot about that. Would explain why travis sees 0.0.0 ☺

matz-e: And travis picks up a better version number. I'd like to get this in and then tag `0.1.0`, if that's ok.

mgeplf: yup, sounds good

PR #43, opened 2019-07-23
Add reports

jorblancoa:
- reportinglib integration
- Added support for serial/parallel implementation


CMake/Catch.cmake
```diff
@@ -0,0 +1,175 @@
+# Distributed under the OSI-approved BSD 3-Clause License.  See accompanying
```
matz-e: /me no likey that we include these files. Can't we grab them from a safe location? Catch should either be found with standard CMake mechanisms or included as a submodule, in which case these files should be referenced from the submodule.

mgeplf: `catch` is already versioned into `libsonata/tests/contrib`

matz-e: Then I think that should be switched to a git submodule to retain some version information.

mgeplf: The author of catch suggests copying the file as was done here: 
https://github.com/catchorg/Catch2/blob/master/docs/tutorial.md#where-to-put-it

jorblancoa: I have seen that copying the catch header file was suggested by the author and most of the people does that, so I thought that adding it as submodule was overkill. That is also why I just  copied the cmake helper files.

But maybe when you start coping several files is worth adding it as a submodule to get always the latest changes.

matz-e: > But maybe when you start coping several files is worth adding it as a submodule to get always the latest changes.

This. Alternatively, I think it would make sense to keep all files from catch in the same folder. Do the versions of the header and the CMake files match?

> The author of catch suggests copying the file as was done here:

That section references a blog post from 2011, I think C++/CMake habits have evolved since then. If we're only using C++, then I think just copying the header is fine. Now we've added CMake files, too.

CMakeLists.txt
```diff
@@ -37,6 +40,23 @@ endif()
...
+    include_directories (${MPI_INCLUDE_PATH})
+endif()
+
+find_package(HDF5)
+if (HDF5_FOUND)
+    if (HDF5_IS_PARALLEL)
+        include_directories (${HDF5_INCLUDE_DIRS})
+        set (HDF5_LIBS ${HDF5_LIBRARIES})
+    else()
+        message (STATUS "HDF5 found, but no parallel IO support was found, HDF5 support in replib will be disabled")
```
matz-e:
```suggestion
        message (STATUS "HDF5 found, but no parallel IO support was found, HDF5 support in the reporting library will be disabled")
```

src/reports/CMakeLists.txt
```diff
@@ -0,0 +1,43 @@
...
+        "library/hdf5_writer.cpp"
+        "library/hdf5_writer.hpp")
+
+set(reportinglib_HEADERS
+        "${SONATA_INCLUDE_DIR}/bbp/reports/records.hpp")
+
+set(client_SOURCES
+        "client/main.cpp")
+
+#include_directories(SYSTEM ${spdlog_include_directory})
```
matz-e:
```suggestion
```

CMake/Catch.cmake
mgeplf: We only need to have one file though; it's in a known place, so it can be directly pointed to.  The version number is embedded into it.

What is the benefit of adding a submodule if there is only a single file?

src/reports/library/io_writer.hpp
```diff
@@ -0,0 +1,29 @@
+#pragma once
+#include <vector>
+#include <memory>
+
+enum KindWriter {HDF5, ASCII, MPIIO};
```
matz-e: This could be `enum class IoWriter::Kind`

src/reports/library/records.cpp
```diff
@@ -0,0 +1,75 @@
+#include <bbp/reports/records.hpp>
+#include <iostream>
+
+#ifdef DEBUG
+// Only with debug
```
matz-e:
```suggestion
```
redundant?

src/reports/library/report.hpp
```diff
@@ -0,0 +1,52 @@
+#pragma once
+
+#include <memory>
+#include <string>
+#include <map>
+
+#include "cell.hpp"
+#include "report_format.hpp"
+
+enum KindReport {COMPARTMENT, SOMA, SPIKE};
```
matz-e: See above: `enum class Report::Kind`?

src/reports/library/reportinglib.cpp
```diff
@@ -0,0 +1,195 @@
+#include <dirent.h>
+#include <stdlib.h>
+#include <new>
+#include <iostream>
+#include <stdlib.h>
+
+#include "reportinglib.hpp"
+/*#include <spdlog/sinks/rotating_file_sink.h>
+#include <spdlog/sinks/stdout_color_sinks.h>*/
```
matz-e: Dead code.

tests/reports/CMakeLists.txt
```diff
@@ -0,0 +1,13 @@
+include_directories(SYSTEM
+                    ${catch2_include_directory}
```
matz-e: Where are these defined?

matz-e: BTW, this is better IMO to just link against the Catch target, and it will take care of the include directories.

mgeplf:
> I think a quick merge could be appropriate.

I would prefer to do cleanup instead of a copy: take the pain now.

CMakeLists.txt
mgeplf: 'the reporting library' is an internal name that might not mean something to people using the library.

CMake/Catch.cmake
jorblancoa: > We only need to have one file though; it's in a known place, so it can be directly pointed to. The version number is embedded into it.
> 
> What is the benefit of adding a submodule if there is only a single file?

The problem is that I was using some cmake helper files to add tests, so I also copied them onto the Cmake folder. I agree with @matz-e that at least we should be consistent with the versions.

Another way would be to add the tests manually like it is done currently in libsonata and remove the need of the cmake files completely.

CMakeLists.txt
```diff
@@ -113,6 +134,10 @@ target_include_directories(sonata_static
     PRIVATE $<TARGET_PROPERTY:HighFive,INTERFACE_INCLUDE_DIRECTORIES>
 )
 
+include_directories(${SONATA_INCLUDE_DIR})
+include_directories(${SONATA_SRC_DIR})
```
mgeplf: why is a source directory being added as an include dir?

matz-e:
> > I think a quick merge could be appropriate.
> 
> I would prefer to do cleanup instead of a copy: take the pain now.

Sure. I noticed that this PR is towards a branch, not `master`, hence that comment.

CMakeLists.txt
```diff
@@ -103,6 +123,7 @@ target_compile_definitions(sonata_shared
 target_link_libraries(sonata_shared
     PRIVATE fmt::fmt-header-only
     PRIVATE HighFive
+    ${MPI_CXX_LIBRARIES}
```
mgeplf: Are these `private`?

include/bbp/reports/records.hpp
```diff
@@ -0,0 +1,70 @@
+#pragma once
+
+#include <reports/library/reportinglib.hpp>
+
+/**
+ * Provides a bridge between c-based programs in order to have access to the c++-based objects
+ */
+#ifndef __cplusplus
```
mgeplf: we don't need C support, right?

include/bbp/reports/records.hpp
```diff
@@ -0,0 +1,70 @@
...
+#endif
+#if defined(__cplusplus)
+#include <cstddef>
+extern "C" {
+#endif
+int records_clear();
+/**
+ * Add cell to existing or new report
+ */
+int records_add_report(const char* reportName,
```
mgeplf: why `const char*` instead of `std::string`?

jorblancoa:
> > > I think a quick merge could be appropriate.
> > 
> > 
> > I would prefer to do cleanup instead of a copy: take the pain now.
> 
> Sure. I noticed that this PR is towards a branch, not `master`, hence that comment.

Exactly, the idea was to "copy" the code from gerrit to a branch here so the code could be reviewed through PRs easily. But I think a first cleanup review would be actually good and helpful.

src/reports/CMakeLists.txt
```diff
@@ -0,0 +1,43 @@
+set(reportinglib_SOURCES
+        "library/cell.cpp"
+        "library/cell.hpp"
```
mgeplf: do the `.hpp` files contain sources, or are they just headers?
Seemingly, the convention in the code is headers are `.h`, and `.hpp` files are included into `.cpp` files, but not compiled.

mgeplf:
Cool, I think we're on the same page, then.

src/reports/client/main.cpp
```diff
@@ -0,0 +1,198 @@
+#include <iostream>
```
mgeplf: applications should probably live somewhere higher up in the source tree

src/reports/client/main.cpp
```diff
@@ -0,0 +1,198 @@
...
+
+#ifdef HAVE_MPI
+#include <mpi.h>
+#endif
+
+#include <bbp/reports/records.hpp>
+
+struct Neuron {
+    // some properties
+    int gid;
```
mgeplf: `gid` isn't a concept in sonata, so this will be confusing, I think.  Is this the same thing as a `node_ids`

src/reports/client/main.cpp
```diff
@@ -0,0 +1,198 @@
...
+#include <cmath>
+
+#ifdef HAVE_MPI
+#include <mpi.h>
+#endif
+
+#include <bbp/reports/records.hpp>
+
+struct Neuron {
+    // some properties
```
mgeplf: comment is superfluous

src/reports/client/main.cpp
```diff
@@ -0,0 +1,198 @@
...
+    if (std::fabs(num_steps * dt + tstart - tstop) > 1e-9) {
+        num_steps++;
+    }
+    double t = 0.0;
+    for(int i=0; i<num_steps; i++) {
+        std::cout << "++++Recording data for t = " << t << std::endl;
+        records_nrec(t, compartment_cellids.size(), &compartment_cellids[0], compartment_report);
+        records_nrec(t, soma_cellids.size(), &soma_cellids[0], soma_report);
+
+        std::this_thread::sleep_for(std::chrono::milliseconds(500));
```
mgeplf: why is this needed? Does `records_nrec` look at the time or something?

src/reports/library/ascii_writer.cpp
```diff
@@ -0,0 +1,49 @@
+#include <iostream>
```
mgeplf: Is this a `sonata` thing?  I can't find a specification for this or the 'bin' format, so I'm a bit confused.

src/reports/library/ascii_writer.cpp
```diff
@@ -0,0 +1,49 @@
...
+    m_file << name << std::endl;
+    size_t buffer_size = total_steps * total_compartments;
+    for(int i=0; i<buffer_size; i++) {
+        m_file << buffer[i] << ", ";
+    }
+    m_file << std::endl << std::endl;
+}
+
+void AsciiWriter::write(const std::string& name, const std::vector<int>& buffer) {
+
```
mgeplf: missing?

src/reports/library/ascii_writer.cpp
```diff
@@ -0,0 +1,49 @@
...
+}
+
+void AsciiWriter::write(const std::string& name, const std::vector<uint32_t>& buffer) {
+    m_file << name << std::endl;
+    for(int elem: buffer) {
```
mgeplf: uint32_t?

src/reports/library/ascii_writer.cpp
```diff
@@ -0,0 +1,49 @@
+#include <iostream>
+
+#include "ascii_writer.hpp"
+
+AsciiWriter::AsciiWriter(const std::string& report_name)
+: IoWriter(report_name), m_file(report_name) {}
+
+AsciiWriter::~AsciiWriter() {}
```
mgeplf: = default;

src/reports/library/ascii_writer.cpp
```diff
@@ -0,0 +1,49 @@
...
+        m_file << buffer[i] << ", ";
+    }
+    m_file << std::endl << std::endl;
+}
+
+void AsciiWriter::write(const std::string& name, const std::vector<int>& buffer) {
+
+}
+
+void AsciiWriter::write(const std::string& name, const std::vector<uint32_t>& buffer) {
```
mgeplf: these can probably be templetized

src/reports/library/ascii_writer.cpp
```diff
@@ -0,0 +1,49 @@
...
+
+void AsciiWriter::write(const std::string& name, const std::vector<float>& buffer) {
+
+}
+
+void AsciiWriter::write(const std::string& name, const std::vector<double>& buffer) {
+
+}
+
+void AsciiWriter::close() {
```
mgeplf: Is this necessary?  Should IOWriter be more RAII?  What is the behavior if the writer is closed, and more data is written to it?

src/reports/library/cell.cpp
```diff
@@ -0,0 +1,48 @@
...
+void Cell::add_spike(double *spike_timestamp) {
+    m_spikes.push_back(spike_timestamp);
+}
+
+int Cell::fill_data(float* data, bool cell_to_be_recorded) {
+    // Copy data from cell data structures to buffer
+    int reportingTotal = 0;
+    if(cell_to_be_recorded) {
+        for (auto &elem: m_compartments) {
+            data[reportingTotal++] = (float) (*elem);
```
mgeplf: please no c-style casts

src/reports/library/cell.cpp
```diff
@@ -0,0 +1,48 @@
+#include <stdlib.h>
+#include <new>
```
mgeplf: what is this needed for?

src/reports/library/cell.cpp
```diff
@@ -0,0 +1,48 @@
...
+    // TODO: what should be the id format?
+    m_compartment_id = gid*1000;
+}
+Cell::~Cell() {}
+
+int Cell::get_num_compartments() {
+    return m_compartments.size();
+}
+
+int Cell::get_num_spikes() {
```
mgeplf: size_t?

src/reports/library/hdf5_writer.hpp
```diff
@@ -0,0 +1,36 @@
+#include <vector>
+#include <fstream>
+#include <mpi.h>
+#include "hdf5.h"
```
mgeplf: Is this file supposed to be in the commit?

CMakeLists.txt
```diff
@@ -16,10 +16,13 @@ option(${PROJECT_NAME}_CXX_WARNINGS "Compile C++ with warnings as errors" ON)
 set(CMAKE_MODULE_PATH ${PROJECT_SOURCE_DIR}/CMake)
 set(CMAKE_SKIP_INSTALL_ALL_DEPENDENCY true)
 
-set(CMAKE_CXX_STANDARD 11)
+
+
+set(CMAKE_CXX_STANDARD 14)
```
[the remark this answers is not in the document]
jorblancoa: std::make_unique, but I can always use std::unique_ptr<T>(new T(...)) in case this would cause problems

CMakeLists.txt
```diff
@@ -16,10 +16,13 @@ option(${PROJECT_NAME}_CXX_WARNINGS "Compile C++ with warnings as errors" ON)
...
 set(CMAKE_CXX_STANDARD_REQUIRED ON)
 
 set(SONATA_INCLUDE_DIR ${PROJECT_SOURCE_DIR}/include)
+set(SONATA_SRC_DIR ${PROJECT_SOURCE_DIR}/src)
```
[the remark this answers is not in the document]
jorblancoa: It is there because the tests use header files that are on src directory

CMakeLists.txt
```diff
@@ -37,6 +40,23 @@ endif()
...
+if (MPI_FOUND)
+    set(HDF5_PREFER_PARALLEL "ON")
+    add_definitions(-DHAVE_MPI)
+    include_directories (${MPI_INCLUDE_PATH})
+endif()
+
+find_package(HDF5)
+if (HDF5_FOUND)
+    if (HDF5_IS_PARALLEL)
+        include_directories (${HDF5_INCLUDE_DIRS})
```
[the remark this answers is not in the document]
jorblancoa: sure, thanks for pointing it out!

CMakeLists.txt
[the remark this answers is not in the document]
mgeplf: nope, it's fine, was just wondering.

`make_unique` is much nicer.

CMakeLists.txt
jorblancoa: Maybe "reports will be disabled"?

CMakeLists.txt
jorblancoa: Yup

CMakeLists.txt
jorblancoa: needed by tests

include/bbp/reports/records.hpp
jorblancoa: C exposed API is required because it will be called from coreneuron

include/bbp/reports/records.hpp
jorblancoa: same than previous comment...coreneuron c calls

src/reports/client/main.cpp
jorblancoa: I agree, but on the other hand this is only related to reports...thats why i kept it there
Maybe a folder called apps on src? (src/apps/reports/main.cpp)
I dont know what would be the best to be honest.

src/reports/client/main.cpp
jorblancoa: Thats true, maybe I should change gid by node_id and compartments by elements?

src/reports/client/main.cpp
jorblancoa: Its just to simulate the computational phase between timesteps

src/reports/client/main.cpp
matz-e: Isn't this sort of a supporting tool? Then `libsonata/utils` or so seems more appropriate.

src/reports/library/ascii_writer.cpp
jorblancoa: This was there initially to allow different kind of reports in the reportinglib (ascii was just a proof of concept, binary was for the old binary format), but its true that it might not make sense to keep them anymore

src/reports/library/ascii_writer.cpp
jorblancoa: Yes!

src/reports/library/ascii_writer.cpp
jorblancoa: I thought about that, but found it hard when you need to pass some MPI types to the functions (i.e. H5T_STD_I32LE)
I would need to check the template type to pass one or another variable...

src/reports/library/ascii_writer.cpp
jorblancoa: That's a good point, I thought about that but it was necessary because coreneuron might call flush() at the end, which will write the remaining steps of the buffer to file and close it.

src/reports/library/cell.cpp
jorblancoa: it was used in previous versions, thanks!

src/reports/library/cell.cpp
jorblancoa: True! I will change all of them

src/reports/library/hdf5_writer.hpp
jorblancoa: No, is a requirement of the library-> find_package(HDF5)
I will change to <hdf5.h> for clarity

src/reports/library/io_writer.hpp
jorblancoa: I didnt know about enum class, thanks!

tests/reports/CMakeLists.txt
jorblancoa: Yes, I think the best idea is to have catch as a submodule and then just do
```
add_subdirectory(Catch2)
target_link_libraries(unittests PRIVATE Catch2::Catch2 ...)
```
As suggested in their documentation

CMakeLists.txt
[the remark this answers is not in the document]
mgeplf: This applies to the whole project though, so I'm not sure it's the right way to do this.  

I'm not a CMake expert (can @tristan0x or @matz-e comment?) but a I think a couple options are that the headers search path should only be modified for only the tests portion, or the tests directly reference their location.

CMakeLists.txt
mgeplf: Refer to above

include/bbp/reports/records.hpp
mgeplf: What is the plan for how libsonata will be used by coreneuron?  Why is it constrained to using only C?  I'm asking because it might make more sense to the C compatibility headers to live on the coreneuron side.

If this is a C compatibility header, why is the extension `.hpp`?

src/reports/library/cell.cpp
```diff
@@ -0,0 +1,48 @@
+#include <stdlib.h>
+#include <new>
+#include <iostream>
+#include <stdlib.h>
+
+#include "cell.hpp"
```
jdcourcol: We don't use "Cell" in sonata but "Node"

src/reports/library/cell.cpp
```diff
@@ -0,0 +1,48 @@
...
+
+Cell::Cell(int gid, int vgid): m_gid(gid), m_vgid(vgid) {
+    // Compartment id would be gid+compartment_id.
+    //  i.e: for gid 3 and compartment 42 -> 3042
+    // TODO: what should be the id format?
+    m_compartment_id = gid*1000;
```
jdcourcol: This sounds fragile no ? It assumes that there is always less than 1000 compartments right

src/reports/library/cell.hpp
```diff
@@ -0,0 +1,30 @@
+#pragma once
+
+#include <memory>
+#include <vector>
+
+class Cell {
+  private:
+    int m_gid;
+    int m_vgid;
```
jdcourcol: what is a vgid ?

src/reports/client/main.cpp
mgeplf: Why does that need to be simulated?

I'm confused as to what the purpose of this application is: its name doesn't really help in that regard.  Is it a test? Is it to help debug? Is it to generate fake data?  Maybe if its scope was more clear, it would make sense why we are pausing here.

src/reports/library/hdf5_writer.cpp
```diff
@@ -0,0 +1,204 @@
...
+
+HDF5Writer::~HDF5Writer() {}
+
+void HDF5Writer::configure_group(const std::string& group_name) {
+    hid_t group = H5Gcreate2(m_file, group_name.c_str(), H5P_DEFAULT, H5P_DEFAULT, H5P_DEFAULT);
+    H5Gclose(group);
+}
+
+void HDF5Writer::configure_attribute(const std::string& group_name, const std::string& attribute_name) {
+    std::cout << "Configuring attribute: " << attribute_name << std::endl;
```
jdcourcol: should we have some logging instead ?

src/reports/library/ascii_writer.cpp
mgeplf: Removing as much as possible will make this much more maintainable, so I'm in favour of removing what we don't need.

src/reports/library/ascii_writer.cpp
mgeplf: using templates for the standard types, and having some specializations for exotic types seems like a good way around this.  However, if the `ascii_writer` isn't needed, it's moot.

src/reports/library/cell.cpp
mgeplf: Please make sure that there aren't unnecessary includes in other files; I've think I've seen the `#include <new>` in a few places, possibly because of cut and paste.

src/reports/library/cell.hpp
```diff
@@ -0,0 +1,30 @@
...
+    ~Cell();
+
+    int fill_data(float* data, bool cell_to_be_recorded);
+
+    void add_compartment(double* compartment_value);
+    void add_spike(double* spike_timestamp);
+    int get_num_compartments();
+    int get_num_spikes();
+
+    const std::vector<uint32_t>& get_compartment_ids() { return m_compartment_ids; }
```
jdcourcol: is uint32 enough ? mouse brain is 70B neurons and you multiply in the cpp by 1000.

src/reports/library/hdf5_writer.hpp
mgeplf: Can you use `HighFive` for the hdf5_writer?  That's what this library uses when possible.

Currently the project doesn't build when HDF5_IS_PARALLEL isn't set because `include_directories (${HDF5_INCLUDE_DIRS})` isn't happening.

src/reports/library/hdf5_writer.cpp
```diff
@@ -0,0 +1,204 @@
...
+
+#ifdef HAVE_MPI
+    // Enable MPI access
+    MPI_Info info = MPI_INFO_NULL;
+    H5Pset_fapl_mpio(plist_id, ReportingLib::m_allCells, info);
+
+    // Initialize independent/collective lists
+    m_collective_list = H5Pcreate(H5P_DATASET_XFER);
+    m_independent_list = H5Pcreate(H5P_DATASET_XFER);
+    H5Pset_dxpl_mpio(m_collective_list, H5FD_MPIO_COLLECTIVE);
```
mgeplf: This also requires HDF5_IS_PARALLEL, which isn't guaranteed to exist

src/reports/library/hdf5_writer.hpp
```diff
@@ -0,0 +1,36 @@
+#include <vector>
+#include <fstream>
+#include <mpi.h>
```
mgeplf: is this necessary in the header?

CMakeLists.txt
```diff
@@ -37,6 +40,23 @@ endif()
...
+find_package(MPI)
+if (MPI_FOUND)
+    set(HDF5_PREFER_PARALLEL "ON")
+    add_definitions(-DHAVE_MPI)
+    include_directories (${MPI_INCLUDE_PATH})
+endif()
+
+find_package(HDF5)
+if (HDF5_FOUND)
+    if (HDF5_IS_PARALLEL)
```
mgeplf: We are getting into combinatoric blowup realm here; the final PR will have to include updates to the testing to make sure all the versions are covered.  

Currently I'm seeing
MPI{ON/OFF} / HDF5{ON/OFF} / HDF5_IS_PARALLEL {ON/OFF} ...

pramodk:
Just to add context  (which is not stated in the description of PR by Jorge):

* this should have been draft PR
* this is PR to **temporary branch (reports)**  and not to master

All reviews are good and very helpful but just want to point out that it's WIP and missing lot of stuff (as you indicated) .

src/reports/client/main.cpp
jorblancoa: It is a test/debug application that simulates the way coreneuron would call the reportinglib.
And coreneuron has a record phase where it passes the pointer of its data to the library and a computation phase between steps. This is just trying to emulate that, but is not actually needed.

src/reports/library/cell.hpp
jorblancoa: virtual gid

src/reports/library/hdf5_writer.hpp
jorblancoa: you are right, in the cpp is enough

src/reports/library/cell.cpp
jorblancoa: Yes, this was just a proof of concept, we would have to think in something different

src/reports/library/cell.cpp
jorblancoa: Sure, I am doing a header cleanup

src/reports/library/hdf5_writer.cpp
jorblancoa: Yes, adding logging with spdlog is in the TODO list

src/reports/library/cell.hpp
jdcourcol: There is not such a thing in SONATA anymore. 
There is  "model types" property instead, that specifies that it is a virtual node eventually. 
"A model_type attribute allows nodes to be configured as biophysical, point_neuron, etc. and also virtual one may be provided to specify external (or virtual) nodes that are not explicitly simulated but provide inputs to the network."

src/reports/library/h5typemap.hpp
```diff
@@ -0,0 +1,80 @@
...
+     *  hdf5 suggest hbool_t instead of c++ type but MPI defines c99 _Bool only.
+     *  Are they all compatible? More important is size_t which is platform dependent
+     *  and may not be 64 byte unsigned int all the time. May be better to use uint64_t
+     *  for all internal offsets that we are going to write to hdf5? */
+
+    template <typename T>
+    hid_t get_h5_type(__attribute__((unused)) T v);
+
+    template <>
+    hid_t inline get_h5_type(__attribute__((unused)) bool v) {
```
mgeplf: have you looked at HighFive::AtomicType?

jorblancoa:
Thanks for all your reviews and input! It has been really helpful.

I will close this for now, fix all comments and then reopen it once ready (so that you guys don't have to review multiple times)

PR #44, opened 2019-07-23
make sure numpy is installed before building bindings

mgeplf:
* skipped if on python2.  Also should be noted that older
  versions of setuptools (ex: 20.7.0) in py3 don't use wheels
  for setup_requires, so you get to build numpy from source...


wizmer:
I should do the same hack for MorphIO then I guess ?

PR #45, opened 2019-08-07
Keep PATCH optional for CMake.

matz-e:
Matches the tagging pattern.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -21,9 +21,9 @@ string(REPLACE "\n"
                PY_PKG_VERSION_FULL
                "${PY_PKG_VERSION_FULL}")
 list(GET PY_PKG_VERSION_FULL -1 PY_PKG_VERSION_FULL)
-# keep MAJOR.MINOR.PATCH
+# keep MAJOR.MINOR.PATCH (PATCH being optional)
 string(REGEX
-       REPLACE "([0-9]+\\.[0-9]+\\.[0-9]+).*"
+       REPLACE "([0-9]+\\.[0-9]+(\\.[0-9]+)?).*"
                "\\1"
                SONATA_VERSION
                "${PY_PKG_VERSION_FULL}")


CMakeLists.txt
```diff
@@ -21,9 +21,9 @@ string(REPLACE "\n"
                PY_PKG_VERSION_FULL
                "${PY_PKG_VERSION_FULL}")
 list(GET PY_PKG_VERSION_FULL -1 PY_PKG_VERSION_FULL)
-# keep MAJOR.MINOR.PATCH
+# keep MAJOR.MINOR.PATCH (PATCH being optional)
 string(REGEX
-       REPLACE "([0-9]+\\.[0-9]+\\.[0-9]+).*"
+       REPLACE "([0-9]+\\.[0-9]+(\\.[0-9]+)?).*"
```
wizmer: Shouldn't it default to zero instead ? Is there a difference between v1.2.0 and v1.2 ?

mgeplf:
Which tagging pattern?  Instead, could we make have the tagging pattern always include the PATCH?

matz-e:
Sorry, my mistake. Just trying to purge some stuff from my local tree. I recently had trouble building `libsonata`, but that maybe just my convoluted setup.

matz-e:
BTW, can we merge PRs without reviews and CI?

mgeplf:
As a policy, or in practice?
As a policy, I would prefer if things need to pass CI and get at least a review; in practice, I don't haven't checked what the current settings are.

matz-e:
In practice. I like to keep settings as restrictive as possible to avoid any mistakes (like accidentally pushing to master).

mgeplf:
Yes, I agree about being restrictive.  Let's have a look together at some point, to set it up right.  I'm not sure I have the permissions at the moment, though.

wizmer:
I just had an issue while trying to `pip install` a dev version.
Now that we use setuptools_scm, we end up with versions looking like:
`0.1.dev45+g1ffb02a.d20190816`

However according to https://cmake.org/cmake/help/v3.5/command/project.html the version number must be: `<major>[.<minor>[.<patch>[.<tweak>]]]]` where each of these variable is a number. I do not find anything about cmake and development versions.

matz-e:
> I just had an issue while trying to `pip install` a dev version.
> Now that we use setuptools_scm, we end up with versions looking like:
> `0.1.dev45+g1ffb02a.d20190816`
> 
> However according to https://cmake.org/cmake/help/v3.5/command/project.html the version number must be: `<major>[.<minor>[.<patch>[.<tweak>]]]]` where each of these variable is a number. I do not find anything about cmake and development versions.

I think I had the same issue, hence this PR. Just could not reproduce it.

matz-e:
BTW, I can merge this without a review? And rebase and merge?

wizmer:
I think instead of keeping PATCH optional, it would be better to make it default to 0, no ?

CMakeLists.txt
matz-e: I don't think there's a difference, really. It's all convention.

matz-e:
> I think instead of keeping PATCH optional, it would be better to make it default to 0, no ?

Theoretically. I don't think it's worth the effort. According to the docs, patch versions shouldn't have been removed.

wizmer:
I think I used to have the admin rights on libsonata but I no longer have them...

PR #47, opened 2019-09-11
versioning: use git directly from CMake

matz-e:
Would make our spack live easier. Should be equivalent to the old version detection.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -1,30 +1,34 @@
 cmake_minimum_required(VERSION 3.3)
 cmake_policy(VERSION 3.3)
 
-execute_process(COMMAND git describe --tags
-                WORKING_DIRECTORY ${CMAKE_SOURCE_DIR}
-                RESULT_VARIABLE GIT_VERSION_FAILED
-                OUTPUT_VARIABLE GIT_PKG_VERSION_FULL
-                ERROR_VARIABLE GIT_VERSION_ERROR
-                OUTPUT_STRIP_TRAILING_WHITESPACE)
-if(GIT_VERSION_FAILED)
-  message(
-    FATAL_ERROR
-      "Could not retrieve version from command 'git describe --tags'\n"
-      ${GIT_VERSION_ERROR})
-endif()
+if(VERSION_INFO)
+    set(SONATA_VERSION ${VERSION_INFO})
+else()
+    execute_process(COMMAND git describe --tags
+                    WORKING_DIRECTORY ${CMAKE_SOURCE_DIR}
+                    RESULT_VARIABLE GIT_VERSION_FAILED
+                    OUTPUT_VARIABLE GIT_PKG_VERSION_FULL
+                    ERROR_VARIABLE GIT_VERSION_ERROR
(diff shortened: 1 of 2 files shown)


CMakeLists.txt
```diff
@@ -2,31 +2,29 @@ cmake_minimum_required(VERSION 3.3)
 cmake_policy(VERSION 3.3)
 
 # get version from Python package
```
mgeplf: out of date

CMakeLists.txt
```diff
@@ -2,31 +2,29 @@ cmake_minimum_required(VERSION 3.3)
...
-find_package(PythonInterp REQUIRED)
-execute_process(COMMAND ${PYTHON_EXECUTABLE} setup.py --version
-                WORKING_DIRECTORY ${CMAKE_SOURCE_DIR}
-                RESULT_VARIABLE PY_VERSION_FAILED
-                OUTPUT_VARIABLE PY_PKG_VERSION_FULL
-                ERROR_VARIABLE PY_VERSION_ERROR
+execute_process(COMMAND git describe --tags
```
mgeplf: how does this work when one is between versions?
Also, are tags guaranteed to be sorted by date?

matz-e: From the docs:
```
The command finds the most recent tag that is reachable from a commit.
```

The current implementation also strips any additional information besides the last commit. CMake does not seem to understand the concept of a development version?

mgeplf: > From the docs

Sadly, git is a moving target, and I've they don't really guarantee stability except for --porcelain commands.  I have the scars to prove it :(

CMake has that deficiency afaik - at least on the python side we can call things `.dev0`

matz-e: I've done some development by Travis… locally checking that CMake does reasonable stuff. In the last incarnation, `git` should not be called when building with `setup.py`, and could be overwritten by the user.

Should we make the pure C++ build fool-proof, too? Since one could just add `-DSONATA_VERSION`, I feel like the C++ build is reasonable.

mgeplf: That seems like nice compromise.

I'm going to add @tomdele @wizmer to the review, so they are in the loop.

matz-e:
> As long as the C++ and python version of tagged commit are the same, I'm okay with that.

Yes, `setup.py` will pre-set the variable, versions should be identical.

CMakeLists.txt
```diff
@@ -1,32 +1,34 @@
...
-               "${PY_PKG_VERSION_FULL}")
-list(GET PY_PKG_VERSION_FULL -1 PY_PKG_VERSION_FULL)
+    string(REPLACE "\n"
+                   ";"
+                   GIT_PKG_VERSION_FULL
+                   "${GIT_PKG_VERSION_FULL}")
+   list(GET GIT_PKG_VERSION_FULL -1 SONATA_VERSION)
+endif()
+
 # keep MAJOR.MINOR.PATCH (PATCH being optional)
```
mgeplf: What do people think about making 'patch' not be optional, and setting the CMake's `<tweak>` to be the count of commits from the since the tag?

PR #50, opened 2019-10-16
Adding move ctors

ferdonline:
Libsonata is implemented using the pimpl idiom. Even though this idiom is know to have limitations, like being non-copyable, it does allow for move constructors. In fact move-constructing is extremely cheap since the only thing moved is the pointer.
A bit all around I see clients of this library using shared pointers due to the inability of moving.
This patch fixes that without any shortcoming.

**Implementation**
The move constructor has to be explicitly defined since the opaque pointer is const. Since data itself is non-const we can safely const_cast the pointer to move it.

**Tests**
A simple test was added where we move NodePopulation.

### DIFF
--- a/include/bbp/sonata/population.h
+++ b/include/bbp/sonata/population.h
@@ -207,6 +207,8 @@ protected:
 
     Population(const Population&) = delete;
 
+    Population(Population&&);
+
     virtual ~Population();
 
     struct Impl;
@@ -225,6 +227,8 @@ public:
     PopulationStorage(const std::string& h5FilePath, const std::string& csvFilePath = "");
 
     PopulationStorage(const PopulationStorage&) = delete;
+
+    PopulationStorage(PopulationStorage&&);
 
     ~PopulationStorage();
 
--- a/include/bbp/sonata/population.h
+++ b/include/bbp/sonata/population.h
@@ -207,9 +207,9 @@ protected:
 
     Population(const Population&) = delete;
 
-    Population(Population&&);
+    Population(Population&&) noexcept;
 
-    virtual ~Population();
+    virtual ~Population() noexcept;
 
     struct Impl;
     const std::unique_ptr<Impl> impl_;
@@ -228,9 +228,9 @@ public:
 
     PopulationStorage(const PopulationStorage&) = delete;
 
-    PopulationStorage(PopulationStorage&&);
+    PopulationStorage(PopulationStorage&&) noexcept;
 
-    ~PopulationStorage();
+    ~PopulationStorage() noexcept;
 
     /**
      * List all populations.
--- a/src/population.cpp
+++ b/src/population.cpp
@@ -145,11 +145,11 @@ Population::Population(
 }
 
 
-Population::Population(Population&& p)
+Population::Population(Population&& p) noexcept
     : impl_(std::move(const_cast<std::unique_ptr<Population::Impl>&>(p.impl_))) {}
 
 
-Population::~Population() = default;
+Population::~Population() noexcept = default;
 
 
 std::string Population::name() const
--- a/src/population.cpp
+++ b/src/population.cpp
@@ -145,6 +145,10 @@ Population::Population(
 }
 
 
+Population::Population(Population&& p)
+    : impl_(std::move(const_cast<std::unique_ptr<Population::Impl>&>(p.impl_))) {}
+
+
 Population::~Population() = default;
 
 
--- a/.travis.yml
+++ b/.travis.yml
@@ -15,11 +15,11 @@ matrix:
       python: 2.7
 
     - os: osx
-      osx_image: xcode10.1
+      osx_image: xcode10.3
       language: generic
       python: 3.6
     - os: osx
-      osx_image: xcode10.1
+      osx_image: xcode10.3
       language: generic
       python: 2.7
 
--- a/src/population.hpp
+++ b/src/population.hpp
@@ -198,14 +198,14 @@ PopulationStorage<Population>::PopulationStorage(const std::string& h5FilePath, 
 
 
 template <typename Population>
-PopulationStorage<Population>::PopulationStorage(PopulationStorage<Population>&& st)
+PopulationStorage<Population>::PopulationStorage(PopulationStorage<Population>&& st) noexcept
     : impl_(std::move(
         const_cast<std::unique_ptr<PopulationStorage<Population>::Impl>&>(st.impl_)))
 {}
 
 
 template<typename Population>
-PopulationStorage<Population>::~PopulationStorage() = default;
+PopulationStorage<Population>::~PopulationStorage() noexcept = default;
 
 
 template<typename Population>
--- a/src/population.hpp
+++ b/src/population.hpp
@@ -197,6 +197,13 @@ PopulationStorage<Population>::PopulationStorage(const std::string& h5FilePath, 
 }
 
 
+template <typename Population>
+PopulationStorage<Population>::PopulationStorage(PopulationStorage<Population>&& st)
+    : impl_(std::move(
+        const_cast<std::unique_ptr<PopulationStorage<Population>::Impl>&>(st.impl_)))
+{}
+
+
 template<typename Population>
 PopulationStorage<Population>::~PopulationStorage() = default;
 
--- a/tests/test_nodes.cpp
+++ b/tests/test_nodes.cpp
@@ -167,3 +167,10 @@ TEST_CASE("NodePopulation", "[base]")
     CHECK(population._dynamicsAttributeDataType("dparam-Y") == "int64_t");
     CHECK(population._dynamicsAttributeDataType("dparam-Z") == "string");
 }
+
+
+TEST_CASE("NodePopulationMove", "[base]") {
+    NodePopulation population("./data/nodes1.h5", "", "nodes-A");
+    NodePopulation pop2 = std::move(population);
+    CHECK(pop2.name() == "nodes-A");
+}


mgeplf:
Could you just make the changes to move ctors, and not apply for the formatting?  That will be done in a separate PR, otherwise it's too much at once.

.clang-format
```diff
@@ -1,15 +1,91 @@
...
+AllowShortIfStatementsOnASingleLine: false
+AllowShortLoopsOnASingleLine: false
+AlwaysBreakBeforeMultilineStrings: true
+AlwaysBreakTemplateDeclarations: true
 BasedOnStyle: WebKit
+BinPackArguments: false
+BinPackParameters: false
 BraceWrapping:
-  AfterClass: true
+  AfterClass: false
```
mgeplf: We spent a bunch of time defining what we wanted for style, so I think it's best if we leave the decisions we had in place.

tristan0x: > We spent a bunch of time defining what we wanted for style, so I think it's best if we leave the decisions we had in place.

FYI this `AfterClass: false` comes from BBP standards, but I have no objection since every project is free to define its own conventions.

ferdonline:
I believe the style in place was not even complete regarding what was decided in those meetings... and anyway It hasn't been applied since long, there were lines well over 100 chars which are highly unfriendly.
That said, I will create different PR, but I think we should homogenise styles

src/population.cpp
```diff
@@ -145,6 +145,10 @@ Population::Population(
 }
 
 
+Population::Population(Population&& p)
+    : impl_(std::move(const_cast<std::unique_ptr<Population::Impl>&>(p.impl_))) {}
```
mgeplf: Hmmm, I find this funny.  Since impl_ is a `unique_ptr`, I'm not sure what the benefit of making it const is, and then casting away the const in this case.

include/bbp/sonata/population.h
```diff
@@ -226,6 +228,8 @@ class SONATA_API PopulationStorage
 
     PopulationStorage(const PopulationStorage&) = delete;
 
+    PopulationStorage(PopulationStorage&&);
```
mgeplf: iirc, @tristan0x showed that move ctors need to be marked `noexcept` for some optimizations to kick in.  Maybe he can comment?

ferdonline: That's a good point (I believe he mentioned the knows case of std::vector which will not move in case it's no except, also for the destructor I believe...)
In this case we don't have copy-ctor so that eventual hidden penalty is not there.
Either way I'd gladly mark it as noexcept. But can we ensure so?

src/population.cpp
ferdonline: I'm just not changing what was already there. With a good compiler I would expect it to compute a fixed offset to the Impl struct, and then we indeed have faster code, so I wouldn't discard.
const_cast is 0 overhead and in this specific case should be safe.

include/bbp/sonata/population.h
ferdonline: Cool! Optimizations in!

src/population.cpp
ferdonline: > I'm not sure I follow. Are you saying the const is necessary for faster code?
Const may lead to faster code, no guarantees since compiler does its own optimizations, right? In this case I don't see any benefit of having it non-const, and a potential benefit.

>  but the pointer can actually be reset w/ a move, meaning that it's not actually const.
It is const so any attempt to modify it will break the "protocol" and lead to an inconsistent object. Here we lie in the shadow of the fact that the object is gonna be destroyed anyway. Any other situation I would be strongly against.

mgeplf: > In this case I don't see any benefit of having it non-const, and a potential benefit.

The benefit of having it non-const is that we're not deceiving the library consumers that something is const when it is not.  I'm also not convinced that this could provide a benefit; I would like to see an example where it happens.

> It is const so any attempt to modify it will break the "protocol" and lead to an inconsistent object.

Which is why I don't want to cast away const, because that is 'breaking the protocol'.  Having the member variable be non-const means that we don't have this inconsistency - the type would behave more naturally - there isn't an exception that 'this thing is const in all cases, except for when you move'.

In my opinion, when writing a library, we have to do our best to steer the users away from potential problems.  However, I'm not a day-to-day cpp guy, so maybe someone who uses it more can weigh in - maybe @tristan0x?

tristan0x: > > In this case I don't see any benefit of having it non-const, and a potential benefit.
> 
> The benefit of having it non-const is that we're not deceiving the library consumers that something is const when it is not. I'm also not convinced that this could provide a benefit; I would like to see an example where it happens.

If you agree with me that the object given to a move constructor is already in the death corridor, we can say that `Population::impl_` member variable is constant during the lifetime of a `Population` instance. In this regard, on the library consumer perspective, `Population::impl_` unique pointer should be marked const. `const` variables allow compilers to make assumptions that allow them to perform optimizations. In that regards we should use const as much as possible.

On the other side, the move semantic needs by design to be able to rip of the object before it gets deleted, which explains the required `const_cast`.

In most cases a `const_cast` removing the constness leads to aliasing problem where an additional non-const pointer points to the same object. It becomes much harder for compilers to keep these object in the register caches.

But in this particular case, the rvalue reference is about to be destroyed, so no aliasing problem in  theory. Recent compilers must (I assume) take a particular attention to such move semantic situations.

So I +1 @ferdonline contribution. I would use `const std::unique_ptr<const Impl> impl_` though.

mgeplf: >If you agree with me that the object given to a move constructor is already in the death corridor, we can say that Population::impl_ member variable is constant during the lifetime of a Population instance. 

The problem is that the site at which something might be moved into isn't obvious from the user's perspective, so they can be mislead by the impl_ being declared const when it isn't _actually_ const.

> const variables allow compilers to make assumptions that allow them to perform optimizations.

I'm having a hard time seeing how this works for member variables, when the const can be cast'd away.  Can you guys make a godbolt showing these optimizations?

ferdonline: > I would use const std::unique_ptr<const Impl> impl_ though.

I am not sure about this one. It is undefined behavior to const_cast a pointer to a location that is const (source: stack overflow, so more official info would be welcome).
Now, in this case the location is in the heap, so probably it doesn't fall in that case. 
In either way, the dptr object could potentially change its fields, right? So not sure if it should be const.

tristan0x: > > If you agree with me that the object given to a move constructor is already in the death corridor, we can say that Population::impl_ member variable is constant during the lifetime of a Population instance.
> 
> The problem is that the site at which something might be moved into isn't obvious from the user's perspective, so they can be mislead by the impl_ being declared const when it isn't _actually_ const.

I think see your point: a class having a move constructor with only const member variables is inconsistent. Furthermore `Population` instance can always be `const` (there are no non-const member functions). Therefore in theory, the move constructor is never invoked in this case.
The most important benefit of having a const private member variable is to prevent shooting yourself in the foot in the implementation. Const or not, it won't make the code faster.
In that regard, I am ok to remove the constness of `impl_` to have a more consistent API, even if the implementation is less safe.

mgeplf: > It is undefined behavior to const_cast a pointer to a location that is const

That's another can of worms why I'm afraid to do operations like this.  I'm also not sure the standardeze would make the distinction wrt heap/stack about UB.

> I think see your point: a class having a move constructor with only const member variables is inconsistent

Yes, that is exactly what I'm trying to say: it's a contradiction.  I wish I could have conveyed it more clearly.

>  Const or not, it won't make the code faster.

Agreed - I don't see how the compiler can perform magic b/c the member variable is const, *if* the const-ness can also be casted away - any assumption the compiler can make to perform an optimization cannot hold.

> In that regard, I am ok to remove the constness of impl_ to have a more consistent API, even if the implementation is less safe.

I would argue that it's actually _more_ safe, because of the UB possibility mentioned above.

ferdonline: > Agreed - I don't see how the compiler can perform magic b/c the member variable is const, if the const-ness can also be casted away - any assumption the compiler can make to perform an optimization cannot hold.

My understanding is the following:
 - const pointer to const location, the compiler might not use a pointer altogether and therefore is UB to modify something that doesn't exist.
 - const pointer to a writable object: the pointer exists but the compiler can assume the object is always the same, thus dereferencing the pointer fewer times.
 - using a const_cast in the latter case is not preventing optimizations, but lifting the consistency checks the user requested.
- we are discussing peanuts, but it's a fun conversation anyway :)

If you have more insight / counter-cases in these points I'd be glad to know. 
Having a pro-example might not be possible since compilers are evolving and might not take advantage of that const yet.

mgeplf: > const pointer to a writable object: the pointer exists but the compiler can assume the object is always the same, thus dereferencing the pointer fewer times.

A const pointer just means that it always points at the same address - the compiler *won't* be able to make any assumptions or dereference things fewer times, so I don't think that's a valid optimization.

> we are discussing peanuts, but it's a fun conversation anyway :)

I disagree, we have to figure out if this is UB - because that's not something that should be merged into the library.  From https://en.cppreference.com/w/cpp/language/const_cast:

`Modifying a const object through a non-const access path ... results in undefined behavior.`

In this case, the std::unique_ptr is being modified in the 'source' population, so I think we should remove the const.

ferdonline: > A const pointer just means that it always points at the same address - the compiler won't be able to make any assumptions or dereference things fewer times

There's a contradiction here. It can indeed assume the address won't change.

> In this case, the std::unique_ptr is being modified in the 'source' population

This is a real fact, and the actual issue here, I agree. Even though the object will exist and updates should be fine, apparently there may be optimizations bringing data race issues in multithreaded setups (https://isocpp.org/blog/2012/12/you-dont-know-const-and-mutable-herb-sutter) and so it seems to still fall in the UB dark cloud :(
I will create a patch to fix. Thanks for bringing this to our attention

mgeplf: > There's a contradiction here. It can indeed assume the address won't change.

Right, but you said `the pointer exists but the compiler can assume the object is always the same, thus dereferencing the pointer fewer times.`  The compiler *cannot* assume the object is the same, it can only assume the *address* is the same, and thus it still has to deference it every time - ie: there won't be a reduction in the deref count.

> I will create a patch to fix.

Thanks.

PR #52, opened 2019-10-18
Reformatting based on coding conventions

ferdonline:
respecting some projects own options, e.g. afterClass break

### DIFF
--- a/.clang-format
+++ b/.clang-format
@@ -1,15 +1,91 @@
-﻿---
+---
+AccessModifierOffset: -2
+AlignAfterOpenBracket: Align
+AlignConsecutiveAssignments: false
+AlignConsecutiveDeclarations: false
+AlignEscapedNewlinesLeft: true
+AlignOperands: true
+AlignTrailingComments: true
+AllowAllParametersOfDeclarationOnNextLine: true
+AllowShortBlocksOnASingleLine: false
+AllowShortCaseLabelsOnASingleLine: false
+AllowShortFunctionsOnASingleLine: Empty
+AllowShortIfStatementsOnASingleLine: false
+AllowShortLoopsOnASingleLine: false
+AlwaysBreakBeforeMultilineStrings: true
+AlwaysBreakTemplateDeclarations: true
 BasedOnStyle: WebKit
+BinPackArguments: false
+BinPackParameters: false
 BraceWrapping:
   AfterClass: true
   AfterControlStatement: false
-  AfterEnum: true
-  AfterFunction: true
+  AfterEnum: false
+  AfterExternBlock: false
+  AfterFunction: false
   AfterNamespace: false
-  AfterStruct: true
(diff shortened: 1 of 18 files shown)


mgeplf:
I was going to make an issue about this; before reformatting everything, I think we need to:
1) say what version of clang format we are going to use (because the output changes over time, so we need to have a stable one)
2) have a CI step that uses that checks the format, so that it matches whatever style is required.

ferdonline:
We can come up with some schemes, but to my mind I see as adequate the code check of the diff.
That way we can have this initial "all-round" reformat without staying tied to a specific version forever.

ferdonline:
@mgeplf It would be nice to include this in the new tag and release in the November "pack"

mgeplf:
I think we can tag a release without this going in: the functionality has been added, so that's what's necessary.

I want to avoid constant churn from reformats, and only have to do this when new 'clang-format's are chosen.

If we don't have a CI check, we'll run into the problem where we periodically have to reformat everything, which makes `git blame` less effective.  If we have a CI check, the version has to be known.  For instance, when I run clang-format on this PR, v4 gives errors with the format config, and 8 & 9 both give some differences - hence the need for stabilization.

ferdonline:
Maybe I failed to express my idea.
In the CI we can check for the style of the modified chunk only. Like that it's more flexible regarding format versions.

pramodk:
> I want to avoid constant churn from reformats, and only have to do this when new 'clang-format's are chosen.
>  For instance, when I run clang-format on this PR, v4 gives errors with the format config, and 8 & 9 both give some differences - hence the need for stabilization.

I agree with this. We should "stabilise" clang-format version to avoid the changes.

IIRC, clang-format 7 is being used in hpc-coding-convention repo but if there is a need to upgrade, we should do it before applying such changes to other projects. We are in same boat for coreneuron project.

nrnhines:
I'm also interested in the idea of a consistent format for NEURON (although the effect on the code
will be very far reaching). But, also in the NEURON context, I worry about the extra difficulty for
contributors to jump through all the hoops involved in pushing an ideal change. i.e.documentation
both for user and developer, proper format, a test that fails without the changeset and succeeds with
it. Anyway, what are the downsides to formatting a pull request via a github hook when submitted
rather than failing the CI?

ferdonline:
As a fact I'm not opposed to reformatting automatically a modified chunk. 

I just see it a bit nonsense to automatically reformat everything (can end up in massive changes) or, even more difficult to maintain, request the user to submit a patch where all the code is already perfectly formatted.

ferdonline:
Can someone change the permissions of the project to require at least one approving review? I can't.

mgeplf:
I have added the required checks in #57, including using clang 9 for formatting checks.  To make sure it worked, I re-formated the code there.  If someone wants to reuse that github action, feel free.

PR #53, opened 2019-10-18
Using non-const pointer to impl

ferdonline:
After the long conversation in #50 we should not const_cast

### DIFF
--- a/src/population.cpp
+++ b/src/population.cpp
@@ -145,8 +145,7 @@ Population::Population(
 }
 
 
-Population::Population(Population&& p) noexcept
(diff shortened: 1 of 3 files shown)


src/population.cpp
```diff
@@ -145,8 +145,7 @@ Population::Population(
 }
 
 
-Population::Population(Population&& p) noexcept
-    : impl_(std::move(const_cast<std::unique_ptr<Population::Impl>&>(p.impl_))) {}
+Population::Population(Population&& p) noexcept = default;
```
mgeplf: nice

PR #56, opened 2019-11-14
improve continuous integration

mgeplf:
* move scripting from .travis.yaml to actual shell scripts
* setup.py on python 3.4 uses the correct number of cores available
  on the machine, otherwise 2 is used so less chance of freezing while
  compiling
* gitignore *.pyc

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -95,7 +95,11 @@ class CMakeBuild(build_ext, object):
                 cmake_args += ["-A", "x64"]
             build_args += ["--", "/m"]
         else:
-            build_args += ["--", "-j"]
+            try:
+                CPU_COUNT = os.cpu_count()  # python 3.4+
+            except:
+                CPU_COUNT = 2
+            build_args += ["--", "-j%d" % CPU_COUNT]
 
         env = os.environ.copy()
         env["CXXFLAGS"] = '{} -DVERSION_INFO=\\"{}\\"'.format(
--- a/setup.py
+++ b/setup.py
@@ -107,7 +107,7 @@ class CMakeBuild(build_ext, object):
                 cmake_args += ["-A", "x64"]
             build_args += ["--", "/m"]
         else:
-            build_args += ["--", "-j".format(max(MIN_CPU_CORES, get_cpu_count()))]
+            build_args += ["--", "-j{}".format(max(MIN_CPU_CORES, get_cpu_count()))]
 
         env = os.environ.copy()
         env["CXXFLAGS"] = '{} -DVERSION_INFO=\\"{}\\"'.format(
--- a/setup.py
+++ b/setup.py
@@ -1,6 +1,5 @@
 import inspect
 import os
-import os.path as osp
 import platform
 import re
 import subprocess
@@ -14,6 +13,7 @@ from distutils.version import LooseVersion
 
 
 REQUIRED_NUMPY_VERSION = "numpy>=1.12.0"
+MIN_CPU_CORES = 2
 
 
 class lazy_dict(dict):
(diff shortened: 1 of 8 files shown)


setup.py
```diff
@@ -95,7 +95,11 @@ def build_extension(self, ext):
                 cmake_args += ["-A", "x64"]
             build_args += ["--", "/m"]
         else:
-            build_args += ["--", "-j"]
+            try:
+                CPU_COUNT = os.cpu_count()  # python 3.4+
```
matz-e: Docs suggest `len(os.sched_getaffinity(0))` instead…

mgeplf: cool; now I need to know why - I assume it interacts better w/ `cgroups`; wonder what it does on the mac

mgeplf: Doc says `They are only available on some Unix platforms.  For more detailed information, consult your Unix manpages.`

I'll look more into it.

matz-e: Can we just remove the concurrency? It's not like `libsonata` is a very large library that absolutely needs a parallel build.

mgeplf: We could remove concurrency, but it's pretty easy to have, and it does make install faster (13s vs. 17s on my old & slow machine)

I just wanted to fix it so it doesn't start *all* the compiles at the same time - `make -j` isn't very smart.

matz-e: Seems like not that much gain…

mgeplf: 30% for 3 lines of code?  Seems like a no effort win to me, but if it's a no go on your end, let me know and I'll remove it.

setup.py
```diff
@@ -95,7 +95,11 @@ def build_extension(self, ext):
...
+            except:
+                CPU_COUNT = 2
+            build_args += ["--", "-j%d" % CPU_COUNT]
```
matz-e:
```suggestion
            build_args += ["--", "-j{}".format(CPU_COUNT)]
```

setup.py
matz-e: Well, it's 30%, but also just 4 seconds. I guess the proper way would be to first test for `sched_getaffinity`, then for `cpu_count`, and then fall back to not having any parallelism? Also safeguard for the fact that any of these methods may just return 0…

Seems like this could be a feature of setuptools by now…

mgeplf: When you're building the set of packages w/ `spack`, how does it manage parallelism? Does it build the packages serially? How many packages are you building? Because 4 seconds probably adds up over time if they are build serially.

Also, at some point this repo will have the reporting stuff in it, afaik, so I'm not sure it's worth removing the parallelism, to potentially put it back in when the compile time increases.

Finally, the main point of this change is to remove the bare `make -j`, which can launch way too many processes.  Whether I use exactly the number of CPUs, or a smaller subset, the behavior is better than saturating the scheduler, and better than compiling things serially, IMO.

matz-e: IIRC, spack calls `make`/`cmake` with some user-specified parallelism (36 for us). There are bigger axes to grind than pure build times in our setup, though.

Anyways, I'm not against building in parallel, but more lamenting that this is not better integrated in `setuptools` and the like. If we end up doing it like I outlined above, we should probably comb through our repositories and fix other instances (i.e., MVDTool).

setup.py
```diff
@@ -95,13 +107,13 @@ def build_extension(self, ext):
                 cmake_args += ["-A", "x64"]
             build_args += ["--", "/m"]
         else:
-            build_args += ["--", "-j"]
+            build_args += ["--", "-j%d" % max(MIN_CPU_CORES, get_cpu_count())]
```
matz-e:
```suggestion
            build_args += ["--", "-j{}".format(max(MIN_CPU_CORES, get_cpu_count()))]
```

mgeplf: For single args, I find the `%` easier to read.

PR #58, opened 2019-11-28
added high resolution logo

mgeplf:


matz-e:
Shouldn't SONATA be all caps or all lowercase?

jdcourcol:
> Shouldn't SONATA be all caps or all lowercase?

indeed. it should be all lowercase here.

mgeplf:
ok, all lower case it will be, once I get the new version

matz-e:
Also, can we get this in a vector format?

mgeplf:
all lowercase:
https://github.com/BlueBrain/libsonata/blob/3c8e8e58b602fdf06190cddea9d6678dc5af98c0/README.md

jdcourcol:
I think there should be 5 staves on the "music sheet" not 6. Or may be it is because there are 6 layers :) ?

mgeplf:
I'm dying here.  Artistic liberty, I think.

matz-e:
The top one could already be just a spike line…

jdcourcol:
> The top one could already be just a spike line…

this is a good suggestion as the current logo does not convey the simulation part. I will talk to Michael.

mgeplf:
I asked for that initially, but apparently it didn't work well.

jdcourcol:
Ok, I asked Michael for an updated logo with a staff removed and eventually some spikes

mgeplf:
ping

matz-e:
svg?

PR #59, opened 2020-01-15
Generate cmake targets to use libsonata as a subproject / submodule

NadirRoGue:


CMakeLists.txt
```diff
@@ -152,6 +152,13 @@ install(TARGETS sonata_shared sonata_static
         DESTINATION lib
 )
 
+install(EXPORT sonata-targets
+        FILE ${PROJECT_NAME}Targets.cmake
+        DESTINATION share/${PROJECT_NAME}/CMake)
+
+export(EXPORT sonata-targets
+       FILE "${PROJECT_BINARY_DIR}/${PROJECT_NAME}Targets.cmake")
+
```
matz-e: How does this differ from line 170? In a regular installation, I see the following CMake files (without this PR):
```
$prefix/share/sonata/CMake/sonata-targets.cmake
$prefix/share/sonata/CMake/sonata-targets-relwithdebinfo.cmake
$prefix/share/sonata/CMake/sonata-config.cmake
```

NadirRoGue: When used as a sub-project, without being installed, these files are not generated in the appropriate location, and neither the config file is generated in the project binary directoy, thus the project that depends on libsonata is unable to locate it

matz-e: How do you include this as a subproject? I was thinking that `add_subdirectory(deps/libsonata)` should already expose the `sonata::…` targets.

NadirRoGue: We are using a CMake framework called CMakeCommon (https://github.com/Eyescale/CMake) that allow us to just specify a github repo and it will automatically download and set it up, however it requires the target files to work

matz-e: When using this as a subproject, I don't think you should call `find_package` later on, but rather use the CMake targets directly?

This essentially duplicates L170, which I'm not a fan of.

NadirRoGue: We will use the `add_subdirectory` approach. Closing pr

PR #60, opened 2020-01-24
method to create a Selection of all members of a population

mgeplf:


src/population.cpp
```diff
@@ -142,6 +142,10 @@ uint64_t Population::size() const {
     return dset.getSpace().getDimensions()[0];
 }
 
+Selection Population::selectAll() const {
```
matz-e: One more empty line before?

Also, couldn't this be inlined?

.travis.yml
```diff
@@ -23,21 +23,27 @@ matrix:
...
       python: 2.7
 
+addons:
+  homebrew:
+    packages:
+    - hdf5
+    update: true
+  apt:
+    packages:
+    - libhdf5-dev
```
matz-e: \o/

mgeplf: Do you have experience with this? I'm surprised it makes things slower.

src/population.cpp
mgeplf: Good call, and I suppose (although I trust the compiler :))

.travis.yml
matz-e: No, just seems a lot better than having to "script" this…

src/population.cpp
mgeplf: Hmm, on second thought, none of the other methods are inlined yet, so I'd prefer to do that in another PR, and not split one out currently.

.travis.yml
mgeplf: Turning off the 'update' makes it faster, based on a sample size of one.

src/population.cpp
matz-e: I think just the non-`impl_` concerning methods should be inlined, right?

.travis.yml
wizmer: What if you keep the Homebrew cache ?

mgeplf: There are a bunch of them, and the process for handling it seems error prone:
https://stackoverflow.com/a/53331571

I'd prefer to use what travis gives automatically, since they probably know best what works/ doesn't work, and I don't want to spend hours debugging it.

src/population.cpp
mgeplf: Would be worth checking if GCC w/ LTO is already doing that, just for fun.

I don't think it's something that would be rejected if someone makes a PR, assuming it isn't super intrusive.  I just don't have time to do it now.

PR #62, opened 2020-01-29
Make bbp::sonata::Selection comparable

### LINKED ISSUE #61: Make `bbp::sonata::Selection` comparable
Primarily to make unit tests easier, but can be useful for other things as well.

mgeplf:
* Closes #61

### DIFF
--- a/include/bbp/sonata/population.h
+++ b/include/bbp/sonata/population.h
@@ -50,6 +50,9 @@ class SONATA_API Selection
     const Ranges ranges_;
 };
 
+bool SONATA_API operator==(const Selection&, const Selection&);
+bool SONATA_API operator!=(const Selection&, const Selection&);
+
 template <typename Iterator>
 Selection Selection::fromValues(Iterator first, Iterator last) {
     Selection::Ranges ranges;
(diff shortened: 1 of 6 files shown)


include/bbp/sonata/population.h
```diff
@@ -50,6 +50,9 @@ class SONATA_API Selection
     const Ranges ranges_;
 };
 
+bool SONATA_API operator==(const Selection&, const Selection&);
+bool SONATA_API operator!=(const Selection&, const Selection&);
```
matz-e: Can we also have `operator bool`?

tristan0x: It is in the Python bindings, but it feels weird to me to override this operator in C++. One should use the `empty()` method like for STL containers: `if (!my_vector.empty())`.

mgeplf: we already have `empty()` like in std::vector

mgeplf: Or what @tristan0x said.

matz-e: Maybe I just like my C++ code to feel more Pythonic…

mgeplf: I try and follow the principle of least astonishment, and in the C++ community, my read is that containers have an `empty()` and having a bool operator would be surprising.

matz-e: …least astonishment…C++… that's a good one. Anyways, fine with me, I don't have strong feelings here.

tristan0x: Thank you for your comment anyway @matz-e

PR #64, opened 2020-01-30
Add comparison for node attributes to values

mgeplf:
* This is in preparation for node_sets resolution

* fixed nodes1.h5 generation & file: enum library didn't have the same
    number of rows as the rest of the dataset
* use Catch2 as a submodule, since I needed the update version for the
  fancy TEMPLATE_TEST_CASE()


python/bindings.cpp
```diff
@@ -336,7 +336,31 @@ PYBIND11_MODULE(libsonata, m) {
...
+            [](NodePopulation& obj, const std::string& name, const std::string& value) {
+                return obj.matchAttributeValues(name, value);
+            },
+            "name"_a,
+            "value"_a,
+            "Return selection where the attribute name has values matching value")
+        .def(
+            "match_values",
+            [](NodePopulation& obj, const std::string& name, const int value) {
+                return obj.matchAttributeValues<int>(name, value);
```
mgeplf: Anyone have a good idea on how to avoid the repetition of `match_values` w/ `int` and `double`?

src/nodes.cpp
```diff
@@ -23,6 +25,91 @@ NodePopulation::NodePopulation(const std::string& h5FilePath,
...
+    }
+
+    // normal attribute
+    return _matchAttributeValues<std::string>(population, name, value);
+}
+
+}  // anonymous namespace
+
+template <typename T>
+Selection NodePopulation::matchAttributeValues(const std::string& name, const T value) const {
```
mgeplf: I'm not really happy w/ this method of dispatch.  Any suggestions are very welcomed.

Basically, I need `typename T` to be comparible to whatever the `HighFive::AtomicType` is - which is dynamic from whatever is loaded.  Currently, I think this might give surprising results if the value is a `float` of 1.5, but the attribute column type is int -> a rounding could happen and you'd get a Selection that is surprising.  Or maybe I'm seeing ghosts.

mgeplf:
Looking for feedback, especially on the comments I posted.  Thanks.

mgeplf:
common! I want a critical eye on this - some of the code strikes me as ugly, and I feel like there are better ways (maybe even some that require `c++14`...)

src/nodes.cpp
matz-e: I think that's a good concern. Should at least be tested. I guess that `_matchAttributeValues` should be called with the HighFive-determined datatype?

It would be a big ugly dispatch table @tristan0x may have a more elegant solution.

tests/test_nodes.cpp
```diff
@@ -150,3 +150,37 @@ TEST_CASE("NodePopulationSelectAll", "[base]") {
...
+                   int8_t,
+                   uint8_t,
+                   int16_t,
+                   uint16_t,
+                   int32_t,
+                   uint32_t,
+                   int64_t,
+                   uint64_t) {
+    NodePopulation population("./data/nodes1.h5", "", "nodes-A");
+    auto sel = population.matchAttributeValues<TestType>("attr-Y", 23);
```
matz-e: Github is messing with my view.

But the data here should be generated with:
```
    attrs.create_dataset('attr-Y', data=np.arange(21., 27., 0.1), dtype=np.int64)
```
For the float/double test-cases.

mgeplf: `dtype=np.int64`?

matz-e: ah, yes, that was left as an exercise for the test implementation. But now that you mention in, line 25 there reads:
```
    attrs.create_dataset('attr-Y', data=np.arange(21., 27.), dtype=np.int64)
```

Should be fixed, too?

python/bindings.cpp
wizmer:
```python
template <typename ValueType, typename NodePopulationClass> void def_match_values(NodePopulationClass &node){
    node.def(
        "match_values",
        [](NodePopulation& obj, const std::string& name, const ValueType& value) {
            return obj.matchAttributeValues(name, value);
        },
        "name"_a,
        "value"_a,
        "Return selection where the attribute name has values matching value");
}
```

and then

```python

    auto nodePopulation = bindPopulationClass<NodePopulation>(m, "NodePopulation", "Collection of nodes with attributes");
    def_match_values<double>(nodePopulation);
    def_match_values<int>(nodePopulation);
    def_match_values<std::string>(nodePopulation);
```

python/test.py
```diff
@@ -161,6 +161,19 @@ def test_enumeration_values(self):
...
+        # string
+        self.assertEqual(self.test_obj.match_values("attr-Z", "bb").flatten().tolist(),
+                         [1])
+
+        # enum
+        self.assertEqual(self.test_obj.match_values("E-mapping-good", "C").flatten().tolist(),
+                         [0, 2, 4, 5])
+
+        # numeric
+        for type_ in (float, int, ):
```
wizmer: `type_` is not used.

src/nodes.cpp
wizmer: Actually even `float` to `float` comparison seems dangerous. Wouldn't it be better to not use `float` with `matchAttribteValues` at all but define a selector based on `operator<` for `float` ?

mgeplf: So, this is in support of node_sets, which only define equality comparisons.  I considered only doing integer comparisons, but I can (barely) see where floating point comparisons could be useful, so I added those too.

python/test.py
mgeplf: thanks

python/bindings.cpp
mgeplf: Did basically that.  Still doesn't feel elegant, though.

tests/test_nodes.cpp
mgeplf: I'm not following. `data=np.arange(21., 27., 0.1)` would create more entries than there are supposed to be.  This is why I had to change `E-mapping-good` above, so that it has six elements.  I have fixed the `21.` issue.

src/nodes.cpp
wizmer: It feels unreliable to me, or at least use a pre-set tolerance so that people know what to expect.

python/bindings.cpp
wizmer: Agreed, but I have nothing better.

python/test.py
```diff
@@ -161,6 +161,19 @@ def test_enumeration_values(self):
...
+            self.assertEqual(self.test_obj.match_values("attr-Y", type_(23)).flatten().tolist(),
+                             [2, ])
```
wizmer: Test it also with a float with many digits

src/nodes.cpp
mgeplf: I don't think having a tolerance is the best plan, especially since this will be called based off the node_set json definition - ie: not by someone reading the docstring.

python/test.py
mgeplf: Since it's an exact comparison, I'm not sure what that is testing, could you elaborate as to why?

wizmer: Because using an int as a float is only testing an edge case. And also because I don't know what would happen if you would save an irrational number on disk (let's say, `sqrt(2)`) and try to find it back afterwards.

mgeplf: It's just bits, so it would compare equal if the bits match, exactly how an represent-able integer gets compared.  Ie; double(sqrt(2.)) is (represented as) `0xcd3b7f669ea0f63f`, and double(2) is `0x0000000000000040`.

tests/test_nodes.cpp
matz-e: It doesn't have the same range. Basically making sure that type resolution works by creating a bunch of values that would get truncated to the same integer.

python/bindings.cpp
```diff
@@ -336,7 +346,11 @@ PYBIND11_MODULE(libsonata, m) {
 
         .def("__ne__", &bbp::sonata::operator!=, "Compare selection contents are not equal");
 
-    bindPopulationClass<NodePopulation>(m, "NodePopulation", "Collection of nodes with attributes");
+    auto node_population = bindPopulationClass<NodePopulation>(
+        m, "NodePopulation", "Collection of nodes with attributes");
+    defMatchValues<int>(node_population);
+    defMatchValues<double>(node_population);
```
matz-e: Now that @wizmer commented on it, I can't unsee this. Can we remove the match on floating point numbers? I agree that minimizes the risk of things going the wrong way…

src/nodes.cpp
matz-e: I think not supporting float conversions then. I thought also that H5 converts types under the hood?

If something is written to disk as `float` and implicitly converted to `double`, will the representation be equal?

mgeplf: > I think not supporting float conversions then. I thought also that H5 converts types under the hood?

You mean double -> float? Or just not having float compare at all?  The latter was my original 'solution' to the whole problem, as I'm not convinced of the utility of it, however the spec says that `node_sets` can have `json number` -> `H5T_IEEE_*LE, H5T_STD_*LE`. So I feel boxed in.

> If something is written to disk as float and implicitly converted to double, will the representation be equal?

You mean if a float is widened to a double? Yes, they are the same (as per conv.fpprom in c++ standardeze).

src/nodes.cpp
```diff
@@ -102,6 +102,10 @@ INSTANTIATE_TEMPLATE_METHODS(uint32_t)
 INSTANTIATE_TEMPLATE_METHODS(int64_t)
 INSTANTIATE_TEMPLATE_METHODS(uint64_t)
 
+#ifdef __APPLE__
+INSTANTIATE_TEMPLATE_METHODS(size_t)
+#endif
```
matz-e: I feel like I've done something like this somewhere already…

mgeplf: yeah, good old Apple.

PR #66, opened 2020-02-05
add coverage ci tests

mgeplf:
* fixed missing coverage
* switch to lcov, since otherwise '50% is the new 100% coverage':
    gcovr uses a metric that includes exceptions, so unless you
    get bad_alloc to throw from every, say, std::string, you won't have
    full coverage... (see Niels Lohmann (of json c++ library fame)
    dealing with this:
    https://stackoverflow.com/questions/46575061/how-to-improve-branch-coverage-in-c


mgeplf:
I'm hoping to get it to 100%, but right now it's this TODO that needs coverage, and I can't think of a way to get it to fail:
https://coveralls.io/builds/28539033/source?filename=src/edge_index.cpp#L209

PR #70, opened 2020-02-06
Add node_ids function to NodePopulation

tomdele:
From the specs, node ids (or label) can be different from the positional
indices. See:
https://github.com/AllenInstitute/sonata/blob/master/docs/SONATA_DEVELOPER_GUIDE.md#nodes---optional-reserved-attributes

The mapping between the positional indices and label indices is done via
the "node_id" field if this one is present in the population. If the
field is not present then: positional indices == label indices which is
the case in our circuits today but may change one day.

With complete circuits, this is important to have access to this field
since source/target_node_id returns label indices and not the positional
indices.


mgeplf:
> With complete circuits, this is important to have access to this field
since source/target_node_id returns label indices and not the positional
indices.

We're not doing this anywhere, right?  If we are, we need to look into it and fix it, I believe.

include/bbp/sonata/nodes.h
```diff
@@ -44,6 +44,8 @@ class SONATA_API NodePopulation: public Population
      *
      * Note: This does not match Dynamics_params datasets
      */
+    std::vector<NodeID> nodeIDs(const Selection& selection) const;
```
matz-e: I think the comment note above relates to the function below.

src/nodes.cpp
```diff
@@ -20,12 +20,28 @@ namespace sonata {
...
+
 NodePopulation::NodePopulation(const std::string& h5FilePath,
                                const std::string& csvFilePath,
                                const std::string& name)
     : Population(h5FilePath, csvFilePath, name, ELEMENT) {}
 
 
+std::vector<NodeID> NodePopulation::nodeIDs(const Selection& selection) const {
+    HDF5_LOCK_GUARD
+    if (H5Lexists(impl_->h5Root.getId(), NODE_ID_DSET, H5P_DEFAULT) > 0) {
```
matz-e: Here's no HighFive function for this? Could also be a `try … catch …`

tomdele: Indeed there is a function in HighFive. I missed it. 
Changed.

include/bbp/sonata/nodes.h
tomdele: Whoo my bad.

src/nodes.cpp
matz-e: Thank you!

tomdele:
I am throwing out some ideas and thoughts here I am thinking while writing so it could be messy: 

- Using only positional ids means we need special supports if we want to support node_sets with the "node_id" mechanism in libsonata. 

- In snap, I will use the label indices and not the positional indices for the queries. So querying with snap and libsonata will returns different values.  

- In libsonata, we do not have any "circuit vision". We just see the Node/EdgePopulations as stand-alone. So we do not have this source/target_node to node population rebound problem. But this is something I am dealing with in snap (and why I needed this feature). 

- I do not see a lot of usecases with libsonata where you ask explicitly for a specific node label (except for the node_set). Usually, the node ids you want to use are coming from a query and the returned ids can be label or positional you don't really care.

- We should probably add a mapping mechanism that takes care of the positional to label indices conversions.

tomdele:
But I think this is for a different review anyway.

mgeplf:
Ahhhhhh, ok, I get what is going on here.

> Using only positional ids means we need special supports if we want to support node_sets with the "node_id" mechanism in libsonata.`

So you interpret this as saying an _explicit_ node_id column is needed there is `node_id = [1,2,3,4]` in the `node_sets`.  I was going to interpret that query as applying to the implicit (positional) indices unless the `node_id` attribute exists - and continue to not support the `node_id` column.  Would that be another option?

tomdele:
We will think a bit more about whether or not we should add this to the lib. It is not quite clear right now.

tomdele:
We decided to close this review and not supporting the `node_id` field.

mgeplf:
See the discussion at https://github.com/AllenInstitute/sonata/issues/3 for more information.

tomdele:
Thanks for the link @mgeplf !

PR #71, opened 2020-02-21
Add output reader for Spikes, somas and elements

alkino:


include/bbp/sonata/output.h
```diff
@@ -0,0 +1,49 @@
...
+
+namespace bbp {
+namespace sonata {
+class SpikesReader {
+  public:
+    SpikesReader(const std::string& filename, const std::string& populationName);
+
+    std::vector<std::pair<uint64_t, double>> get() const;
+    std::vector<std::pair<uint64_t, double>> get(uint64_t node_id) const;
+    std::vector<std::pair<uint64_t, double>> get(std::vector<uint64_t> node_ids) const;
```
mgeplf: As per ticket, please use the `libsonata::Population::Selection` mechanism.

include/bbp/sonata/output.h
```diff
@@ -0,0 +1,49 @@
...
+}  // unnamed namespace
+
+namespace bbp {
+namespace sonata {
+class SpikesReader {
+  public:
+    SpikesReader(const std::string& filename, const std::string& populationName);
+
+    std::vector<std::pair<uint64_t, double>> get() const;
+    std::vector<std::pair<uint64_t, double>> get(uint64_t node_id) const;
```
mgeplf: I'm tempted to say that this and `std::vector<std::pair<uint64_t, double>> get(double tend) const;` aren't necessary - people can call `get(Selection{id})` and `get(/* tstart */ 0, double tend);` respectively.

However, I'm open to keeping them, depending on what others think (@tomdele @matz-e?)

include/bbp/sonata/output.h
```diff
@@ -0,0 +1,49 @@
...
+#include <vector>
+
+namespace {
+template <class InputIt1, class InputIt2, class OutputIt, class BinaryOperation, class BinaryPredicate>
+OutputIt transform_if(InputIt1 first1, InputIt1 last1, InputIt2 first2, OutputIt d_first, BinaryOperation binary_op, BinaryPredicate binary_pred) {
+    while (first1 != last1) {
+        if (binary_pred(*first1, *first2)) {
+            *d_first++ = binary_op(*first1, *first2);
+        }
+        first1++, first2++;
```
mgeplf: pre-increment to avoid the copy?  Also, please put them on separate lines. (I may need to override the `operator,`  some day /sarcasm)

alkino:
Are you more happy with this c++ API?

mgeplf:
Wow, fast.

First glance looks great; let me play with it a bit, and hopefully get feedback from others, but I think you nailed it.

alkino:
`python3 -c "import libsonata; s = libsonata.SpikesReader('/home/cornu/libsonata/tests/reports/integration/spikes.ref'); print(s['All'].get(libsonata.Selection([34])))"`
> [(34, 0.2625), (34, 0.2625), (34, 0.2625), (34, 0.2625), (34, 0.2625), (34, 0.2625), (34, 0.2625), (34, 0.2625), (34, 0.2625), (34, 0.2625), (34, 0.2625), (34, 0.2625), (34, 0.2625), (34, 0.2625), (34, 0.2625), (34, 0.2625), (34, 0.2625), (34, 0.2625), (34, 0.2625), (34, 0.2625), (34, 0.2625), (34, 0.2625), (34, 0.2625), (34, 0.2625), (34, 0.2625), (34, 0.2625), (34, 0.2625), (34, 0.2625), (34, 0.2625), (34, 0.2625), (34, 0.2625), (34, 0.2625), (34, 0.2625), (34, 0.2625), (34, 0.2625), (34, 0.2625), (34, 0.2625), (34, 0.2625), (34, 0.2625), (34, 0.2625), (34, 0.2625), (34, 0.2625), (34, 0.2625), (34, 0.2625), (34, 0.2625), (34, 0.2625), (34, 0.2625), (34, 0.2625), (34, 0.2625), (34, 0.2625), (34, 0.2625), (34, 0.2625), (34, 0.2625), (34, 0.2625), (34, 0.2625), (34, 0.2625), (34, 0.2625), (34, 0.2625), (34, 0.2625), (34, 0.2625), (34, 0.2625), (34, 0.2625), (34, 0.2625), (34, 0.2625), (34, 0.2625), (34, 0.2625), (34, 0.2625), (34, 0.2625), (34, 0.2625), (34, 0.2625)]

alkino:
Is there no better way to have a `Selection`?

alkino:
Giving a try to have `print(s['All'])` returning all spikes data

mgeplf:
> Is there no better way to have a Selection in python?

Depends what you're looking to do:
https://github.com/BlueBrain/libsonata/#selection

At some point we'll support the DSL for the node_sets spec, but at the moment that happens at a higher level.  What would qualify as `better` for you?

alkino:
> > Is there no better way to have a Selection in python?
> 
> Depends what you're looking to do:
> https://github.com/BlueBrain/libsonata/#selection
> 
> At some point we'll support the DSL for the node_sets spec, but at the moment that happens at a higher level. What would qualify as `better` for you?

By adding `py::implicitly_convertible<py::list, Selection>();`, I can now use `spikes.get([4,5])`, instead of `spikes.get(libsonata.Selection([4,5]))`

include/bbp/sonata/output.h
```diff
@@ -0,0 +1,95 @@
...
+        ++first1, ++first2;
+    }
+    return d_first;
+}
+}  // unnamed namespace
+
+namespace bbp {
+namespace sonata {
+/**
+  const SpikesReader file(filename);
```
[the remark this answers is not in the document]
alkino: Do you have an idea?

include/bbp/sonata/output.h
```diff
@@ -0,0 +1,97 @@
...
+      std::cout << "[" << timestamp << "] " << node_id << std::endl;
+  }
+*/
+class SONATA_API SpikeReader
+{
+  public:
+    class Population
+    {
+      public:
+        std::vector<std::pair<uint64_t, double>> get() const;
```
tomdele: `NodeID` from common instead of the uint64_t?

include/bbp/sonata/output.h
```diff
@@ -0,0 +1,97 @@
...
+        std::vector<std::pair<uint64_t, double>> get(Selection node_ids) const;
+        std::vector<std::pair<uint64_t, double>> get(double tstart, double tend) const;
+
+      private:
+        explicit Population(const std::string& filename, const std::string& populationName);
+
+        std::vector<uint64_t> node_ids;
+        std::vector<double> timestamps;
```
tomdele: Maybe having these vectors sorted by increasing timestamp could be beneficial afterwards.
This could allow some optimizations for:
`std::vector<std::pair<uint64_t, double>> get(double tstart, double tend) const;`
And would make the outputs more "readable".

What do you think @matz-e , @mgeplf ?

tomdele:
I think this is really nice overall !

include/bbp/sonata/output.h
jorblancoa: I don't think this should be the case, since the SONATA spike file has this attribute called sorting, that sorts the node_ids/timestamps by_time or by_id
https://github.com/AllenInstitute/sonata/blob/master/docs/SONATA_DEVELOPER_GUIDE.md#output-file-formats

So I guess the users should be able to read the values in the order they were written.

tomdele: hum ok I see.
So it could be a good Idea to give access to this `sorting` field so the consumer don t have to resort the output again if not needed.

src/output.cpp
```diff
@@ -26,29 +26,28 @@ auto SpikeReader::operator[](const std::string& populationName) const -> const P
     return getPopulation(populationName);
 }
 
-std::vector<std::pair<NodeID, double>> SpikeReader::Population::get() const {
-    std::vector<std::pair<NodeID, double>> vec;
+SpikeReader::Population::Spikes SpikeReader::Population::get() const {
+    std::vector<Spike> vec;
```
jorblancoa: Spikes?

include/bbp/sonata/output.h
alkino: Miss this part of the specification, I will try to do something nice with it.

python/test.py
```diff
@@ -3,7 +3,9 @@
 
 import numpy as np
 
-from libsonata import EdgeStorage, NodeStorage, Selection, SonataError
+from libsonata import EdgeStorage, NodeStorage, Selection, SonataError, SpikeReader
+
+from libsonata import SpikeReaderPopulation
```
matz-e: What about `SpikePopulation`? A little easier to read. Also, can we either get all imports on one line, or every import on a slngle one?

alkino: In fact, `SpikeReaderPopulation` is an internal python type. Nobody should care about it, it is here only for test.
I don't know what you think about it. People cannot create it (no public ctor).

matz-e: Still, I think the `Reader` part is a little superfluous here. `SpikeReader` returning a `SpikePopulation` makes it a little easier than having to think about what a `SpikeReaderPopulation` is and how it differs from just a population of spikes?

mgeplf: Suppose there is a writing API, would it be the same object as `SpikePopulation`, or would it be different?

alkino: Could be the same.

include/bbp/sonata/output.h
alkino: Done

include/bbp/sonata/output.h
alkino: Done

include/bbp/sonata/output.h
alkino: Done

include/bbp/sonata/output.h
alkino: `SpikeReaderPopulation` => `SpikePopulation`

alkino: By the way, what should I do about sorting?

jorblancoa: I believe nothing! This should come sorted from the writing side. Maybe just check if it is sorted according to the enum value?

tomdele: I think checking if the values are correctly sorted is definitely a good idea. 
I don t know how big the reports will be but thanks to the sorting, you can avoid looping over all gids/timestamps everytime you do a get. But I guess if it s fast enough you can skip it.

alkino: I find the checking useless, if data are corrupted, it's life...

I will try to make some smart search.

alkino: I find the checking useless, if data are corrupted, it's life...

I will try to make some smart search.

mgeplf: > I find the checking useless, if data are corrupted, it's life...

Agreed - if `sorting` says it's sorted, we have to take that at face value - checking if it's sorted is expensive.

python/bindings.cpp
```diff
@@ -443,5 +446,91 @@ PYBIND11_MODULE(libsonata, m) {
...
+        .def_readonly("data", &DataFrame::data)
+        .def_readonly("index", &DataFrame::index);
+    py::class_<SpikeReader::Population>(m, "SpikePopulation", "A population inside a SpikeReader")
+        .def(
+            "get",
+            [](const SpikeReader::Population& self) { return self.get(); },
+            "Return all spikes")
+        .def(
+            "get",
+            [](const SpikeReader::Population& self, double tstart, double tend) {
```
jorblancoa: Consistency with changes on cpp API
```suggestion
            [](const SpikeReader::Population& self, double tstart, double tstop) {
```

python/test.py
```diff
@@ -225,5 +225,59 @@ def test_select_all(self):
...
+        self.assertRaises(RuntimeError, self.test_obj.__getitem__, 'foobar')
+
+    def test_get_reports_from_population(self):
+        self.assertEqual(self.test_obj['All'].times(), (0., 1., 0.1))
+        self.assertEqual(self.test_obj['All'].timeUnits(), 'ms')
+        self.assertEqual(self.test_obj['All'].dataUnits(), 'mV')
+        self.assertTrue(self.test_obj['All'].sorted())
+        #self.assertEqual(len(self.test_obj['All'].get()), 200)
+        #self.assertEqual(self.test_obj['All'].get(node_ids=[13, 14], tstart=0.8, tend=1.0).data, [0.8, 0.9])
+        sel = self.test_obj['All'].get(node_ids=[13, 14], tstart=0.8, tend=1.0)
```
jorblancoa:
```suggestion
        sel = self.test_obj['All'].get(node_ids=[13, 14], tstart=0.8, tstop=1.0)
```

include/bbp/sonata/output.h
```diff
@@ -0,0 +1,135 @@
...
+
+namespace H5 = HighFive;
+
+namespace bbp {
+namespace sonata {
+
+struct SONATA_API DataFrame {
+    using DataType = std::map<NodeID, std::vector<float>>;
+    DataType data;
+    std::vector<float> index;
```
jorblancoa: I don't like index as a name, I found it confusing. Maybe current_step? index_tstep? tstep_index?
And it should probably be double (like tstart, tstop, tstep).

include/bbp/sonata/output.h
```diff
@@ -0,0 +1,135 @@
...
+        using Spikes = std::vector<Spike>;
+        Spikes get(const Selection& node_ids = Selection({}),
+                   double tstart = -1,
+                   double tstop = -1) const;
+        Sorting getSorting() const;
+
+      private:
+        Population(const std::string& filename, const std::string& populationName);
+
+        Spikes spikes;
```
jorblancoa: private members with _ at the end for consistency with the rest of the codebase (spikes_)

include/bbp/sonata/output.h
alkino: This is the name of pandas.DataFrame, I only match it.

https://pandas.pydata.org/pandas-docs/stable/reference/api/pandas.DataFrame.html

```python
dataframe = pandas.DataFrame(data=sonata.data, index=sonata.index)
```

alkino: ok for double. I was confuse because data are in float but times are double.

alkino:
@mgeplf Could you please review this quickly?

tomdele:
Checked on python side and I think @alkino corrected all the bugs.

src/output.cpp
```diff
@@ -0,0 +1,386 @@
...
+}
+void SpikeReader::Population::filterNode(Spikes& spikes, const Selection& node_ids) const {
+    if (sorting_ == Sorting::by_id) {
+        filterNodeIDSorted(spikes, node_ids);
+    } else {
+        filterNodeIDUnsorted(spikes, node_ids);
+    }
+}
+
+void SpikeReader::Population::filterNodeIDUnsorted(Spikes& spikes,
```
mgeplf: Ah, I see what you mean be inplace: these are static, so could just be bare functions.  Perhaps they should be?

src/output.cpp
```diff
@@ -0,0 +1,386 @@
...
+
+        {  // Get times
+            std::vector<double> times;
+            mapping_group.getDataSet("time").read(times);
+            tstart_ = times[0];
+            tstop_ = times[1];
+            tstep_ = times[2];
+            mapping_group.getDataSet("time").getAttribute("units").read(time_units_);
+            size_t i = 0;
+            for (double t = tstart_; t < tstop_ - EPSILON; t += tstep_) {
```
mgeplf: couldn't this have accumulation error, depending on `tstep_`?

src/output.cpp
```diff
@@ -0,0 +1,386 @@
...
+        throw std::runtime_error(
+            "In spikes file, 'node_ids' and 'timestamps' does not have the same size.");
+    }
+
+    std::transform(std::make_move_iterator(node_ids.begin()),
+                   std::make_move_iterator(node_ids.end()),
+                   std::make_move_iterator(timestamps.begin()),
+                   std::back_inserter(spikes_),
+                   [](Spike::first_type&& node_id, Spike::second_type&& timestamp) {
+                       return std::make_pair(std::move(node_id), std::move(timestamp));
```
mgeplf: are node_id & timestamp always POD?  don't need moves, then, right?

src/output.cpp
```diff
@@ -0,0 +1,386 @@
...
+
+template <typename T>
+std::pair<T, std::vector<float>> make_value(NodeID node_id,
+                                            uint32_t element_id,
+                                            std::vector<float> values);
+
+template <>
+std::pair<NodeID, std::vector<float>> make_value(NodeID node_id,
+                                                 uint32_t /* element_id */,
+                                                 std::vector<float> values) {
```
mgeplf: do you need the copy?

src/output.cpp
```diff
@@ -0,0 +1,386 @@
...
+            .read(data);
+
+        std::vector<uint32_t> element_ids;
+        pop_group_.getGroup("mapping")
+            .getDataSet("element_ids")
+            .select({it->second.first}, {it->second.second - it->second.first})
+            .read(element_ids);
+        for (size_t i = 0; i < element_ids.size(); ++i) {
+            std::vector<float> data_by_node;
+            for (auto& datum : data) {
```
mgeplf: can we reserve `data_by_node`?

CMakeLists.txt
```diff
@@ -83,6 +83,7 @@ set(SONATA_SRC
     src/hdf5_mutex.cpp
     src/nodes.cpp
     src/population.cpp
+    src/output.cpp
```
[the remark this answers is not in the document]
jorblancoa: I guess @alkino chose this name because of the title on the SONATA specification
https://github.com/alkino/sonata/blob/fix_index_count/docs/SONATA_DEVELOPER_GUIDE.md#output-file-formats

But I agree with you that report_reader(s) makes more sense

include/bbp/sonata/output.h
```diff
@@ -0,0 +1,143 @@
...
+                   double tstop = -1) const;
+        Sorting getSorting() const;
+
+      private:
+        Population(const std::string& filename, const std::string& populationName);
+
+        Spikes spikes_;
+        Sorting sorting_ = Sorting::none;
+        // Use for clamping of user values
+        double tstart_, tstop_;
```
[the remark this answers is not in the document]
jorblancoa: I think that being in the report reader context, this doesn't need to be emphasized, otherwise almost everything should be prefixed with report_

include/bbp/sonata/output.h
```diff
@@ -0,0 +1,143 @@
...
+        friend SpikeReader;
+    };
+
+    explicit SpikeReader(const std::string& filename);
+
+    std::vector<std::string> getPopulationsNames() const;
+
+    const Population& getPopulation(const std::string& populationName) const;
+
+    const Population& operator[](const std::string& populationName) const;
```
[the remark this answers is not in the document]
jorblancoa: I think this was done because in the python bindings users are used to the bracket notation to refer to a population.
I don't think it hurts to have the option also in the c++ api right?

python/bindings.cpp
```diff
@@ -443,5 +501,51 @@ PYBIND11_MODULE(libsonata, m) {
...
+        .def_property_readonly(
+            "sorting",
+            [](const SpikeReader::Population& self) {
+                auto s = self.getSorting();
+                if (s == SpikeReader::Population::Sorting::none)
+                    return "none";
+                if (s == SpikeReader::Population::Sorting::by_id)
+                    return "by_id";
+                if (s == SpikeReader::Population::Sorting::by_time)
+                    return "by_time";
```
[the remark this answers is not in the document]
jorblancoa: I just checked and GCC doesn't throw any warning while Clang does.
We could just return "none" at the end since by default no sorting can be assumed.

include/bbp/sonata/output.h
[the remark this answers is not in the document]
mgeplf: I see your point, I'm just reading the code, and `tstart`/`tend` is used for the desired times of interest, as well as maximum report time, so I thought it might be worth differentiating them.

include/bbp/sonata/output.h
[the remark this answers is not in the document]
mgeplf: I'd prefer to have consistency within the C++ interface for now.  If all populations are going to work like this, then we can have a separate PR updating all of them.  A grey zone where part of the API works one way, and then another works another way is annoying from a user's standpoint.

include/bbp/sonata/population.h
```diff
@@ -48,7 +48,7 @@ class SONATA_API Selection
     bool empty() const;
 
   private:
-    const Ranges ranges_;
+    Ranges ranges_;
```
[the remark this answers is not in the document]
jorblancoa: Because with the const member, the operator= is disabled and this line wouldn't work:
`nodes_ids_ = Selection::fromValues(values);`

jorblancoa:
> Looks great, most (all?) of my comments are nit picks.

I have addressed some of your comments, however others are really specific and it would be better if @alkino himself answers after he comes back from holidays next week.

Since there are nit picks, could we merge this now and let @alkino take care of the rest of the comments in a separate PR? Or do you think something needs to be addressed before that?

mgeplf:
I've made some of the changes, please have a look.

include/bbp/sonata/population.h
```diff
@@ -48,7 +48,7 @@ class SONATA_API Selection
     bool empty() const;
 
   private:
-    Ranges ranges_;
+    const Ranges ranges_;
```
mgeplf: @jorblancoa: keeping this constant means we don't have to worry about `operator=` for now; in fact it pointed out that `nodes_ids_` in `ReportReader<T>::Population::get` was building a `Selection` only to `flatten` it after

jorblancoa:
> I've made some of the changes, please have a look.

Looks good to me!

PR #72, opened 2020-02-25
Update license files to follow standard format

haleepfl:
Update license files to follow the format suggested by https://choosealicense.com/licenses/lgpl-3.0/

Note: we don't currently have the boilerplate in all files, which is recommended.


matz-e:
Hm, I'd be tempted to say that we should keep the separate license file… isn't it very easy to spot what the license is, then?

haleepfl:
Using these file names allows GitHub to detect the license and display it in the header next to the contributors, see [NeuroR](https://github.com/BlueBrain/NeuroR) for example.

Additionally, the naming of the license files is recommended in the licenses themselves (see [GPLv3](https://choosealicense.com/licenses/gpl-3.0/) for example).

If the minimal text is included in the readme, then I'm not sure we're losing anything?

matz-e:
Ah, sorry! Somehow the `COPYING` slipped me! I guess I'm just used to something called `LICENSE`…

Issue #65, opened 2020-01-31
Bump pybind version to 2.3.0+


haleepfl:
pybind v2.3.0 introduced improvements to the mkdoc.py documentation generation tool. It would be nice if we could bump the version and take advantage of generating our python docstrings from our C++ doxygen documentation, to reduce duplication where possible.

PR #73, opened 2020-03-06
Bump pybind11 to v2.5.0

haleepfl:
Closes https://github.com/BlueBrain/libsonata/issues/65


tomdele:
Can you check with @alkino before merging this ?
We really need the report part today and I don t want a bump of version to mess everything ...

haleepfl:
Is it worth bumping to 2.5.0 since it has been released?

mgeplf:
I would say so, yes.

PR #77, opened 2020-03-12
Mac python wheels

mgeplf:


setup.py
```diff
@@ -166,5 +166,8 @@ def run(self):
     zip_safe=False,
     setup_requires=setup_requires,
     install_requires=install_requires,
-    use_scm_version=True,
+    use_scm_version={"local_scheme": "no-local-version",
```
wizmer: Should I use 'no-local-version' in MorphIO as well ?

wizmer:
I had the same issue. I haven't solved it though. It seems that with this implementation, if you `pip install .` then you cannot run the tests from within the root folder. Hence: https://github.com/BlueBrain/MorphIO/blob/master/ci/python_test.sh#L24

However it works if you `pip install -e .`...

mgeplf:
> I had the same issue. I haven't solved it though

I have (basically the same fix).  There isn't really a good solution that I know of.  Idly wondering if pytest blocks file system local imports.

setup.py
mgeplf: maybe - it's so I can upload to the test.pypi.org

Issue #84, opened 2020-04-08
Python Spike/ReportPopulation .get() behavior with empty list of ids


tomdele:
The current behavior of the report's get function is:
```
from libsonata import SpikeReader
spikes = SpikeReader("spikes.h5")["population"]
spikes.get([])  #  returns all the node ids
spikes.get([-1])  #  returns all the node ids
spikes.get([-2])  #  returns empty list
spikes.get(None)  # raises
```

would be great to have instead: 
```
from libsonata import SpikeReader
spikes = SpikeReader("spikes.h5")["population"]
spikes.get([])  #  returns empty list
spikes.get([-1])  #  returns  empty list
spikes.get([-2])  #  returns empty list
spikes.get(None)  # returns all the node ids
```

alkino:
So, once we get a `Selection` c++-side it is not possible to disambiguate `[]` and `None`.

For this `-1` and `-2` weird bug, I remove the implicit conversion from list and tuple because `Selection` take `Ranges` as arguments.

So let's play the old way:
```python
import numpy as np
s = libsonata.SpikeReader("test/spikes.h5")
s['All'].get(libsonata.Selection(np.array([1,2,3])))
```

alkino:
I think this ticket is over, right?

mgeplf:
I think so; do the tests cover the cases in the original issue? Can you find the commit that fixed it, and link it here?

PR #85, opened 2020-04-20
Fix creating selection

alkino:
Trying to do something for #84 
At the end nothing is good.


alkino:
Keep the current behavior.

PR #86, opened 2020-04-21
selection: add a constructor from a vector of values

alkino:


alkino:
Before I make this patch working, what about the idea?

Against?

I don't like this useless "fromValues" method, it adds nothing except indirection.

mgeplf:
> Against?

Nope, looks fine to me.  Such is life that we have to keep backwards compat for now w/ the `Selection::fromValues`.  We just need to make sure there is symmetry with the other Selection creation, where an array of ints gives the 'flat' ids, and an array of tuples is interpreted as Ranges.

tomdele:
It looks good indeed.

alkino:
I think it's too hard to change this.

PR #87, opened 2020-05-07
Use the current source directory to determine version.

matz-e:
Solves issues when used as a submodule, as seen in Brion. Note that this
variable cannot be `PROJECT_SOURCE_DIRECTORY` as `project()` has not yet
been called.


matz-e:
> ...still not a fan of this tags based release

Aware of that… this is more of a CMake pit to fall into, IMO

PR #88, opened 2020-05-12
Create documentation

haleepfl:
Create the documentation (using sphinx and doxygen) for both Python and C++ APIs.

The documentation is built here: https://libsonata.readthedocs.io/en/latest/

A follow up PR will deal with the consolidation of C++ and Python docstrings to reduce duplication.


haleepfl:
>     * The license badge is red on both RTD and the Github README
This is because the license is not currently set on PyPI. This is fixed by adding the license to the setup.py and making a new release.

>     * The header of the logo image repeats the header on RTD, in a way. Looks a little bit weird (I guess your other PR may address that? Although it removes the license file…)
This is just due to the image used, we can modify it if really necessary I suppose.

>     * In the C++ API docs, I can't seem to get the links to anchors to work
Yes, this is an issue that we don't currently have a good solution for. Currently we use an iframe to show the docs generated by doxygen, so the anchors won't work. Alternatively, we can use Breathe to post process the doxygen output into RST, but this also loses some of the content as not everything is supported.

mgeplf:
Very impressive, thanks for doing all this, it's amazing.

>  This is fixed by adding the license to the setup.py and making a new release.

I think you've made the change already, a release just needs to be done, at some point, right?

PR #89, opened 2020-05-13
Use common docstrings between C++ and Python

haleepfl:


ci/check_generated_docstrings.sh
```diff
@@ -6,19 +6,15 @@
 set -euo pipefail
 
 VENV=build/venv-docstrings
-CLANG_FORMAT_VERSION=9.0.0
+CLANG_VERSION=6.0.0.2
```
matz-e, line 9: :cold_face:

haleepfl: pybind11 actually has the required clang bindings as a submodule, so I've removed this

haleepfl:
@matz-e and @mgeplf this is still a work in progress, but before I continue on, I would appreciate any feedback you have. That way I can address any comments as I transition remaining docstrings.

mgeplf:
looks reasonable to me, so far

matz-e:
Sorry, had this open and forgot about it. Looks OK to me!

…that generated macro expansion is just sick though. Makes anything I write in TeX look very sane.

mgeplf:
Wow, again, tonnes of work, thanks.  So far everything looks reasonable, and I like the output I've seen.  Seems about as maintainable as one can expect too.  :+1:

alkino:
I'm not sure we should push generated file to git, make it generated at compile time by cmake.

haleepfl:
> I'm not sure we should push generated file to git, make it generated at compile time by cmake.

We debated this initially. Normally I would agree, however in this case the generation requires clang. Do we want to make that a required dependency just for people to compile it locally? Mitsuba (which also uses this approach) commits the docstring file to git: https://github.com/mitsuba-renderer/mitsuba2/blob/master/include/mitsuba/python/docstr.h

PR #90, opened 2020-05-13
Make output of "python setup.py sdist" installable

mgeplf:
* Catch2 was previously a header only file, but that was
  changed to a submodule.  To save space in the source distribution,
  this isn't needed, so added a SONATA_TESTS CMake option to allow
  building without tests, and hence the Catch2 dependency


genric:
LGTM

PR #91, opened 2020-05-25
Push sdist

mgeplf:
* dev versions of libsonata weren't being pushed to PyPI, allow that
* Note: wheels are still created in separate repo


haleepfl:
Why not just tag master with a dev version and allow travis to handle the upload in the usual way?

mgeplf:
> Why not just tag master with a dev version and allow travis to handle the upload in the usual way?

I'd prefer not to have to tag each commit; this way setuptools_scm uploads a `devX` version for each merged PR, which seems reasonable to me.

.travis.yml
```diff
@@ -36,18 +36,15 @@ cache:
     - $HOME/.cache/pip
 
 before_install:
-  - rm -rf dist *.egg-info && python setup.py sdist
+    - rm -rf dist *.egg-info
 
 install: skip
```
wizmer, line 41: Is this line useful ?

.travis.yml
```diff
@@ -36,18 +36,15 @@ cache:
...
 script:
-  - ci/cpp_test.sh
-  - ci/python_test.sh
+    - ci/cpp_test.sh
+    - ci/python_test.sh
 
 deploy:
+  provider: script
+  script: ci/python_sdist.sh
```
wizmer, line 48: I would use a filename that suggest this script has a side effect. sdist_to_pypi.sh or something like this.

ci/python_sdist.sh
```diff
@@ -0,0 +1,20 @@
...
+# upload the sdist files to PyPI
+
+set -euo pipefail
+
+if [[ -z "$TRAVIS" ]]; then
+    echo "This script is meant to be run on travis-ci.com"
+    exit -1
+fi
+
+env
```
wizmer, line 12:
```suggestion
```

haleepfl:
> > Why not just tag master with a dev version and allow travis to handle the upload in the usual way?
> 
> I'd prefer not to have to tag each commit; this way setuptools_scm uploads a `devX` version for each merged PR, which seems reasonable to me.

I see. Why not just replace `on: tags: true` with `on: branch: master` then?

https://docs.travis-ci.com/user/deployment#conditional-releases-with-on

PR #93, opened 2020-06-04
Fix report_reader to have several time the same index (node_id, compa…

alkino:
…rtment_id)


src/report_reader.cpp
```diff
@@ -371,7 +371,7 @@ DataFrame<T> ReportReader<T>::Population::get(const Selection& selection,
             for (auto& datum : data) {
                 data_by_node.push_back(datum[i]);
             }
-            data_frame.data.insert(make_value<T>(node_id, element_ids[i], std::move(data_by_node)));
+            data_frame.data.push_back(make_value<T>(node_id, element_ids[i], std::move(data_by_node)));
```
cattabiani, line 374: maybe emplace_back could be better

alkino: As `make_value` return a R-value (prvalue to be correct), we don't care.

src/report_reader.cpp
```diff
@@ -359,19 +355,21 @@ DataFrame<T> ReportReader<T>::Population::get(const Selection& selection,
             .select({index_start, it->second.first},
                     {index_stop - index_start + 1, it->second.second - it->second.first})
             .read(data);
+        int index = 0;
+        for (auto& datum : data) {
+            for (auto d : datum) {
+                data_frame.data[index].push_back(d);
+            }
```
tomdele, line 362: I think changing the names would be beneficial for readability. 
I am a bit lost with the indexing order... 

datum is column or line from the initial 2D array ?
index is ids or times ?

tomdele: You can also remove the second auto --> float. This makes clear why you don `&`

src/report_reader.cpp
```diff
@@ -359,19 +355,21 @@ DataFrame<T> ReportReader<T>::Population::get(const Selection& selection,
             .select({index_start, it->second.first},
                     {index_stop - index_start + 1, it->second.second - it->second.first})
             .read(data);
+        int timer_index = 0;
+        for (const std::vector<float>& datum : data) {
+            for (float d : datum) {
+                data_frame.data[timer_index].push_back(d);
+            }
+            ++timer_index;
+        }
```
tomdele, line 364: Sorry I am lost again, in python I have : 
data_frame.data[x][y] with x == nodes and y == times

You have the opposite here. Is this due to the C vs F representation in bindings ?

tomdele: or  is  it `ids_index` instead ?

include/bbp/sonata/report_reader.h
```diff
@@ -18,9 +18,10 @@ namespace sonata {
 // KeyType will be NodeID for somas report and pair<NodeID, uint32_t> for elements report
 template <typename KeyType>
 struct SONATA_API DataFrame {
-    using DataType = std::map<KeyType, std::vector<float>>;
-    std::vector<double> index;
-    DataType data;
+    using DataType = std::vector<KeyType>;
+    std::vector<double> times;
+    DataType ids;
+    std::vector<std::vector<float>> data;
```
tomdele, line 24: once we are sure about the first and second indices for data we can add a comment here:
// data[ids][times] 
or 
// data[times][ids]

src/report_reader.cpp
```diff
@@ -359,19 +355,21 @@ DataFrame<T> ReportReader<T>::Population::get(const Selection& selection,
             .select({index_start, it->second.first},
                     {index_stop - index_start + 1, it->second.second - it->second.first})
             .read(data);
+        int index = 0;
+        for (auto& datum : data) {
```
jorblancoa:
```suggestion
        for (const auto& datum : data) {
```

jdcourcol:
@alkino can this move forward ?

src/report_reader.cpp
alkino: I have no idea, it's  clearly difficult to tell, too many people read things as they want in the middle ^^

src/report_reader.cpp
alkino: This is a float now

src/report_reader.cpp
alkino: In fact if we look at the c++ tests it's clearly [times][ids]

jorblancoa:
LGTM!

tomdele:
I think we still lack quite a lot of testing here. 

There is no tests for the ElementReader ? 
You do not test any data values in the python part also. 

These need to be tested for sure.

src/report_reader.cpp
tomdele: Yes I checked again I messed up this is indeed [times][ids]

alkino:
test are in python for ElementReader, test in cpp are more here to help me develop.
What do you mean by data values?

tomdele:
I think you should test the C++ side intensively also. I guess some people (from viz or elsewhere) would prefer to use the C++ version. Also testing both with the same expected values can detect binding problems.

In the python part, you are testing some length from the data but never the outputted data themselves. There is nothing that guaranty you retrieve the correct data.

tomdele:
Also did you drop the idea of numpy array for the data ?

tomdele:
Also edge cases are missing on the python side : 
- empty list 
- numpy array 
- empty tuple 
- swapping ids : [1, 2] vs [2, 1]
- the out of range times
- the minus -1 ids
etc ... 

Also I found this using elements.h5 : 
```
rep.get(node_ids=[99999]).data                                                                                                                                                                                              
[[], [],  [],  [],  [],  [],  [],  [],  [],  [],  [],  [],  [],  [],  [],  [],  [],  [],  [], []]

rep.get(node_ids=[999999]).times                                                                                                                                                       
[0.0, 0.2, 0.4, 0.6000000000000001, 0.8, 1.0, 1.2, 1.4, 1.5999999999999999, 1.7999999999999998, 1.9999999999999998, 2.1999999999999997, 2.4, 2.6, 2.8000000000000003, 3.0000000000000004, 3.2000000000000006, 3.400000000000001, 3.600000000000001, 3.800000000000001]

rep.get(node_ids=[999999]).ids
[]
```
I am fine with this behavior but is it the one you intended ?

mgeplf:
Two things:

1) What is the functional testing strategy for this?  Should it have beeen able to catch the problem solved by this PR?
2) In the future, could we give the opportunity to the other people who had participated in the  review to give their approval before merge - that way they can check if their satisfied as well.

alkino:
I'm on it.

alkino:
>     1. What is the functional testing strategy for this?  Should it have beeen able to catch the problem solved by this PR?

The problem fixed by this PR is not a bug, this is a new way to see data, so I fix test, because we misunderstood things. So the test check what we things as the good way to see things.

PR #96, opened 2020-06-17
Add more tests

alkino:
Still to do:

- [x] do the same for Spikes
- [x] tests for ReportReader too
- [x] test inverted Selection
- [x] test data in python
- [x] add a function to get node_ids


tests/test_report_reader.cpp
```diff
@@ -4,6 +4,13 @@
 
 using namespace bbp::sonata;
 
+void testTimes(const std::vector<double>& vec, double start, double step, int size) {
+    REQUIRE(size == vec.size());
+    for (int i = 0; i < vec.size(); ++i) {
+        REQUIRE(std::abs(vec[i] - (start + step * i)) < std::numeric_limits<double>::epsilon());
```
tomdele, line 10: so there is no accumulation of error for this one ?

tests/test_report_reader.cpp
```diff
@@ -23,23 +30,89 @@ TEST_CASE("SpikeReader", "[base]") {
...
+TEST_CASE("SomaReportReader limits", "[base]") {
     const SomaReportReader reader("./data/somas.h5");
 
     REQUIRE(reader.getPopulationsNames() == std::vector<std::string>{"All", "soma1", "soma2"});
 
-    REQUIRE(reader.openPopulation("All").getTimes() == std::make_tuple(0., 1., 0.1));
+    auto pop = reader.openPopulation("All");
+
+    // ids out of range
+    REQUIRE(pop.get(Selection({{100, 101}})).ids == DataFrame<NodeID>::DataType());
```
tomdele, line 41: what about : 
```
Selection({{99, 101}})
```
?

tests/test_report_reader.cpp
```diff
@@ -23,23 +30,89 @@ TEST_CASE("SpikeReader", "[base]") {
...
+    REQUIRE_THROWS(pop.get(Selection({{-1, 1}})));
+
+    // Times out of range
+    REQUIRE(pop.get(Selection({{1, 2}}), 100., 101.).ids == DataFrame<NodeID>::DataType());
+    testTimes(pop.get(Selection({{1, 2}}), 100., 101.).times, 0.0, 0.0, 0);
+    REQUIRE(pop.get(Selection({{1, 2}}), 100., 101.).data == std::vector<std::vector<float>>({}));
+
+    // Inverted times
+    REQUIRE(pop.get(Selection({{1, 2}}), 0.2, 0.1).ids == DataFrame<NodeID>::DataType({1}));
+    testTimes(pop.get(Selection({{1, 2}}), 0.2, 0.1).times, 0.2, 0.1, 1);
```
tomdele, line 59: so there is no throw when t_start > t_stop ?

tests/test_report_reader.cpp
```diff
@@ -23,23 +30,89 @@ TEST_CASE("SpikeReader", "[base]") {
...
+    REQUIRE(pop.get(Selection({{1, 2}}), 0.2, 0.1).data ==
+            std::vector<std::vector<float>>({{1.2}}));
+
+    // Negatives times
+    REQUIRE(pop.get(Selection({{1, 2}}), -1., -2.).ids == DataFrame<NodeID>::DataType({1}));
+    testTimes(pop.get(Selection({{1, 2}}), -1., -2.).times, 0.0, 0.1, 10);
+    REQUIRE(pop.get(Selection({{1, 2}}), -1., -2.).data ==
+            std::vector<std::vector<float>>(
+                {{1.0}, {1.1}, {1.2}, {1.3}, {1.4}, {1.5}, {1.6}, {1.7}, {1.8}, {1.9}}));
```
tomdele, line 68: not sure understand why we have values here

tests/test_report_reader.cpp
```diff
@@ -23,23 +30,89 @@ TEST_CASE("SpikeReader", "[base]") {
...
-    REQUIRE(reader.openPopulation("All").get(Selection({{3, 5}}), 0.2, 0.5).ids ==
-            DataFrame<NodeID>::DataType{{3, 4}});
+    REQUIRE(pop.getSorted());
 
-    REQUIRE(
-        reader.openPopulation("All").get(Selection({{3, 5}}), 0.2, 0.5).data ==
-        std::vector<std::vector<float>>{{{3.2f, 4.2f}, {3.3f, 4.3f}, {3.4f, 4.4f}, {3.5f, 4.5f}}});
+    auto data = pop.get(Selection({{3, 5}}), 0.2, 0.5);
+    REQUIRE(data.ids ==
+            DataFrame<std::pair<NodeID, uint32_t>>::DataType{
```
tomdele, line 111: Could it be beneficial to have a :
```
using ElementID = uint32_t;
```
?

tomdele:
> For my part, I consider it as implementation details, so any value is good as you asked for invalid ranges.

Not sure what you mean here.

tests/test_report_reader.cpp
alkino: it's read directly from file

tests/test_report_reader.cpp
alkino: I will add a test

tests/test_report_reader.cpp
alkino: no, we invert them

tests/test_report_reader.cpp
alkino: it means everything

tests/test_report_reader.cpp
alkino: Yes ;)

alkino:
> > For my part, I consider it as implementation details, so any value is good as you asked for invalid ranges.
> 
> Not sure what you mean here.

I mean, do we really care about bad users inputs and add a lot of tests for it in code, or is the user responsible to give bad values?

alkino:
After all, I have no idea about the output for those limit cases, users should define it, I can adapt.

tests/test_report_reader.cpp
tomdele: ok I see

tests/test_report_reader.cpp
tomdele: ok

tests/test_report_reader.cpp
tomdele: ok good. 
Then maybe having tests for : 
```
pop.get(Selection({{1, 2}}), -1., something)

pop.get(Selection({{1, 2}}), 2., -1)
```
can be great.

tests/test_report_reader.cpp
tomdele: Cool. 
I was puzzled by the `uint32_t` and I had to check the specs again ...

tomdele:
> I mean, do we really care about bad users inputs and add a lot of tests for it in code, or is the user responsible to give bad values?

There are 2 things here to me: 
- the bad inputs from the users : we must accept that the user is not necessary an expert in coding or that he has sometimes no idea of what a good selection or time range are. We should help him in finding his error directly through the code usually via proper error throwing or warning or whatever. This will save a us a lot of painful support time at the end. You are a bit protected in HPC but I have in average maybe between 2 and 5 different scientists asking me stuffs everyday about circuit / simulation reading (and we already spent tonnes of hours on testing everything and documenting everything in bluepy or bluepysnap and in most of nse products).
- the implicit behaviors : this needs to be defined somewhere (usually in the doc). A user cannot find by its own all the implicit behaviors and being sure this is not a bug. How can someone knows that there is no simulation with negative times and or that the times are swapped if t_start > t_stop ? And these implicit behaviors are part of the api and should remain the same if the version change (or the change should be mentioned in a change log). 

Having the tests, ensure that the implicit behaviors remain the same with the changes in the implementation.

tomdele:
> After all, I have no idea about the output for those limit cases, users should define it, I can adapt.

I put some of the limit case behaviors here : 
https://github.com/BlueBrain/libsonata/issues/84

For the rest, I think this is more about you choosing what seems reasonable or not. Then if someone complains you can change it.

tomdele:
About the returned `.data` on the python side, it is mandatory to return a numpy array. @WeinaJi  found out that the dataframe constructor with list is utterly bad (I only tried with the numpy arrays back in the days when we discussed this) :

```
# fake element reader like data :
times = [1, 2]
ids = [(i, j) for i in range(600) for j in range(300)]         # 180000 elements                                                                                                                                  
data = np.random.random((2, len(columns)))              # 2*180000 values                                                                                                                                    
data_list = data.tolist()


%timeit pd.DataFrame(data=data_list, columns=pd.MultiIndex.from_tuples(ids), index=times)                                                                                                       
6.68 s ± 42.5 ms per loop (mean ± std. dev. of 7 runs, 1 loop each)

%timeit pd.DataFrame(data=data, columns=pd.MultiIndex.from_tuples(ids), index=times)                                                                                                            
21.7 ms ± 481 µs per loop (mean ± std. dev. of 7 runs, 10 loops each)
```

Also I tried the `.ids` output options with the same data : 
```
ids = [(i, j) for i in range(600) for j in range(300)]
ids_array = np.array([[i, j] for i in range(600) for j in range(300)]).T

%timeit pd.MultiIndex.from_tuples(ids)                                                                                                                                                         
20.6 ms ± 64.6 µs per loop (mean ± std. dev. of 7 runs, 10 loops each)

%timeit pd.MultiIndex.from_arrays(ids_array)                                                                                                                                                   
1.93 ms ± 6.64 µs per loop (mean ± std. dev. of 7 runs, 1000 loops each)
```
The difference is massive. 

So to summarize :
```
times = [1, 2]
ids = [(i, j) for i in range(600) for j in range(300)]        
ids_array = np.array([[i, j] for i in range(600) for j in range(300)]).T                                                                                                                                  
data = np.random.random((2, len(columns)))                                                                                                                                  
data_list = data.tolist()

# current version
%timeit pd.DataFrame(data=data_list, columns=pd.MultiIndex.from_tuples(columns), index=time)                                                                                                       
7.08 s ± 108 ms per loop (mean ± std. dev. of 7 runs, 1 loop each)

# full numpy
%timeit pd.DataFrame(data=data, columns=pd.MultiIndex.from_arrays(columns_array), index=time)                                                                                                      
2.39 ms ± 18.4 µs per loop (mean ± std. dev. of 7 runs, 100 loops each)
```

So basically if you can provide a numpy array for both the data and the ids, the gain in perf on the python / dataframe side will be HUGE.

tests/test_report_reader.cpp
tomdele: ping !

src/report_reader.cpp
```diff
@@ -261,57 +261,49 @@ bool ReportReader<T>::Population::getSorted() const {
...
-    } else {
-        for (const auto& time_index : times_index_) {
-            if (tstart < time_index.second + EPSILON) {
-                indexes.first = time_index.first;
-                break;
-            }
-        }
+
+    double start = tstart.value_or(tstart_);
+    double stop = tstop.value_or(tstop_);
```
tomdele, line 268: const ?

tomdele:
The changes are really nice ! 

Some comments : 
- it would be nice to have the same "optional" API for the spikes
- missing tests similar to `SomaReportReader limits` for `ElementReader`
- the selection that does not raise with negative numbers or out of range ids 
- `unit32_t` --> `ElementID`
- on the python's side it would be nice to throw a `LibsonataError` instead of a generic`RuntimeError`
- add tests to check if get(node_ids=[1, 2]) != get(node_ids=[2, 1]). This is important for me to know if the ids are always ordered or not. This is used when converting to dataframe (I need to know if I should call .sort(axis=1) or not). 
- some tests are disabled for the moment and should be enabled
- numpy outputs (but for another PR probably)

include/bbp/sonata/report_reader.h
```diff
@@ -16,7 +16,7 @@ namespace H5 = HighFive;
 namespace bbp {
 namespace sonata {
 
-// KeyType will be NodeID for somas report and pair<NodeID, uint32_t> for elements report
+// KeyType will be NodeID for somas report and pair<NodeID, Elementid> for elements report
```
tomdele, line 19: Elementid --> ElementID

include/bbp/sonata/report_reader.h
```diff
@@ -127,8 +127,9 @@ class SONATA_API ReportReader
     mutable std::map<std::string, Population> populations_;
 };
 
+using ElementId = uint32_t;
 using SomaReportReader = ReportReader<NodeID>;
-using ElementReportReader = ReportReader<std::pair<NodeID, uint32_t>>;
+using ElementReportReader = ReportReader<std::pair<NodeID, ElementId>>;
```
tomdele, line 132: ElementID for consistency

tests/test_report_reader.cpp
alkino: now it throws

alkino:
> The changes are really nice !
> 
> Some comments :
> 
>     * it would be nice to have the same "optional" API for the spikes
> 
>     * missing tests similar to `SomaReportReader limits` for `ElementReader`
> 
>     * the selection that does not raise with negative numbers or out of range ids
> 
>     * `unit32_t` --> `ElementID`
> 
>     * on the python's side it would be nice to throw a `LibsonataError` instead of a generic`RuntimeError`
> 
>     * add tests to check if get(node_ids=[1, 2]) != get(node_ids=[2, 1]). This is important for me to know if the ids are always ordered or not. This is used when converting to dataframe (I need to know if I should call .sort(axis=1) or not).
> 
>     * some tests are disabled for the moment and should be enabled in the c++ side
> 
>     * in python having tests with the `.data` is needed. You can use : `numpy.testing.assert_allclose` to compare more easily the floats.
> 
>     * also having a function `ReportReader<T>::Population::nodeIds()` (`node_ids()` in python) would be useful. This function should just access the `population/mapping/node_ids` and return a vector from the dataset (numpy array in python).
> 
>     * numpy outputs (but for another PR probably)

First comment updated, trying to summarize with all things asked in this patch.

alkino:
@tomdele inverted Selection is already tested and throw

tomdele:
Sorry @alkino what are you referring to ? 

> @tomdele inverted Selection is already tested and throw

include/bbp/sonata/report_reader.h
```diff
@@ -93,17 +94,19 @@ class SONATA_API ReportReader
         std::string getTimeUnits() const;
         std::string getDataUnits() const;
         bool getSorted() const;
+        std::vector<NodeID> getNodesIds() const;
```
tomdele, line 97: `getNodesIds` --> `getNodeIds` for consistency with the rest of the names.

include/bbp/sonata/report_reader.h
```diff
@@ -93,17 +94,19 @@ class SONATA_API ReportReader
...
+                               const nonstd::optional<double>& _tstop = nonstd::nullopt) const;
 
       private:
         Population(const H5::File& file, const std::string& populationName);
-        std::pair<size_t, size_t> getIndex(double tstart, double tstop) const;
+        std::pair<size_t, size_t> getIndex(const nonstd::optional<double>& tstart, const nonstd::optional<double>& tstop) const;
 
         std::vector<std::pair<NodeID, std::pair<uint64_t, uint64_t>>> nodes_pointers_;
         H5::Group pop_group_;
+        std::vector<NodeID> nodes_ids_;
```
tomdele, line 109: `nodes_ids_` --> `node_ids_`

include/bbp/sonata/report_reader.h
```diff
@@ -126,8 +129,9 @@ class SONATA_API ReportReader
     mutable std::map<std::string, Population> populations_;
 };
 
+using ElementID = uint32_t;
```
tomdele, line 132: shouldn't this be part of common.h ?

python/bindings.cpp
```diff
@@ -315,31 +325,15 @@ void bindReportReader(py::module& m, const std::string& prefix) {
...
         .def("get",
              &ReportType::Population::get,
              "Return reports with all those node_ids between 'tstart' and 'tstop'",
-             "node_ids"_a,
-             "tstart"_a,
-             "tstop"_a)
+             "node_ids"_a = nonstd::nullopt,
+             "tstart"_a = nonstd::nullopt,
+             "tstop"_a = nonstd::nullopt)
+        .def("get_nodes_ids",
```
tomdele, line 334: `get_node_ids` with no `s` to nodes to be consistent with the SpikeReport / FrameReport etc ...

src/report_reader.cpp
```diff
@@ -108,18 +109,25 @@ auto SpikeReader::openPopulation(const std::string& populationName) const -> con
...
 
-Spikes SpikeReader::Population::get(const Selection& node_ids, double tstart, double tstop) const {
-    tstart = tstart < 0 ? tstart_ : tstart;
-    tstop = tstop < 0 ? tstop_ : tstop;
-    if (tstart > tstop_ + EPSILON || tstop < tstart_ - EPSILON || tstop < tstart) {
-        return Spikes{};
+Spikes SpikeReader::Population::get(const nonstd::optional<Selection>& node_ids,
+                                    const nonstd::optional<double>& tstart,
+                                    const nonstd::optional<double>& tstop) const {
+    double start = tstart.value_or(tstart_);
```
tomdele, line 115: const ?

include/bbp/sonata/optional.hpp
```diff
@@ -0,0 +1,1715 @@
+//
```
matz-e, line 1: Such pain. Should this file not be vendored properly?

alkino: how? I can add this to spack but **it** (totally undefined) want to limit number of depends.

alkino:
Is this the final patch?

include/bbp/sonata/report_reader.h
```diff
@@ -118,22 +119,24 @@ class SONATA_API ReportReader
...
          */
         bool getSorted() const;
+        std::vector<NodeID> getNodeIds() const;
 
         /**
          * \param node_ids limit the report to the given selection.
-         * \param tstart return spikes occurring on or after tstart. tstart=-1 indicates no limit.
-         * \param tstop return spikes occurring on or before tstop. tstop=-1 indicates no limit.
+         * \param tstart return spikes occurring on or after tstart. tstart=None indicates no limit.
+         * \param tstop return spikes occurring on or before tstop. tstop=None indicates no limit.
```
tomdele, line 127: should be tstart=nullopt / tstop=nullopt no ?

include/bbp/sonata/report_reader.h
tomdele: So you did not move this to common.h at the end ?
What's the reason behind that ?

python/tests/test.py
```diff
@@ -308,11 +311,9 @@ def test_get_reports_from_population(self):
         self.assertEqual(keys, [(13, 30), (13, 30), (13, 31), (13, 31), (13, 32), (14, 32), (14, 33), (14, 33), (14, 34), (14, 34)])
 
         self.assertEqual(len(sel.times), 3)  # Number of timestamp (0.8, 1.0 and 1.2)
-        sel = self.test_obj['All'].get(tstart=5., tstop=-1)  # tstart out of range
-        self.assertEqual(len(sel.ids), 0)
-        sel = self.test_obj['All'].get(tstart=3., tstop=3.)
-        self.assertEqual(len(sel.ids), 100)
-
```
tomdele, line 315: what happens for `.get(node_ids=[])` ?

src/report_reader.cpp
tomdele: ping

tests/test_report_reader.cpp
```diff
@@ -23,23 +30,105 @@ TEST_CASE("SpikeReader", "[base]") {
...
 }
 
-TEST_CASE("ReportReader", "[base]") {
+TEST_CASE("SomaReportReader limits", "[base]") {
+    const SomaReportReader reader("./data/somas.h5");
+
+    auto pop = reader.openPopulation("All");
+
+    // ids out of range
+    // REQUIRE_THROWS(pop.get(Selection({{100, 101}})));
```
tomdele, line 39: what s the current behavior ?

Same for `pop.get(Selection({{1, 1}}))`

python/tests/test.py
tomdele: Actually the behavior is correct imao. This is just about a test lacking.

tomdele:
The error handling is so much better and the global behavior is much clearer !
Except the minor changes I have pointed out I think this is in a much better shape.
Congrats @alkino

tomdele:
Just for the record : if the spike reports will start to be massive, I think it would be a good idea to return a different object with : 
`res.times` and `res.ids` instead of the list of tuples.

python/tests/test.py
alkino: Today, it takes all nodes. Do you prefer an empty `DataFrame`?

tests/test_report_reader.cpp
alkino: The current behavior is the one from `Selection` ^^

python/tests/test.py
tomdele: > Today, it takes all nodes.
ah my bad ! I forgot I hacked bluepysnap for this ... 
Yes having an empty dataframe is what I want. We already have the None to return all nodes then we need something for the empty dataframe and [] is perfect.

tests/test_report_reader.cpp
tomdele: perfect

alkino: [1, 1] is [1]

src/report_reader.cpp
```diff
@@ -108,18 +109,25 @@ auto SpikeReader::openPopulation(const std::string& populationName) const -> con
...
+                                    const nonstd::optional<double>& tstop) const {
+    const double start = tstart.value_or(tstart_);
+    const double stop = tstop.value_or(tstop_);
+
+    if (start < 0 - EPSILON || stop < 0 - EPSILON) {
+        throw SonataError("Times cannot be negative");
+    }
+
+    if (start >= stop) {
+        throw SonataError("tstart should be < to tstop");
```
WeinaJi, line 123: I notice the change here that it doesn't allow ```tstart == stop```. Then how would users query for one single timestamp?

alkino: hm, you are true.I will remove this =

tests/test_report_reader.cpp
alkino: I make it throw again.

python/tests/test.py
alkino: hm ok

tests/test_report_reader.cpp
alkino: hum `{{1, 1}}` is refused by `Selection`: `Invalid range: 1-1`

tomdele:
Can we have the same `get(node_ids=[])` behavior for the SpikeReport ?

python/tests/test.py
```diff
@@ -308,11 +311,9 @@ def test_get_reports_from_population(self):
...
+        with self.assertRaises(SonataError): self.test_obj['All'].get(tstart=5.)  # tstart out of range
+        with self.assertRaises(SonataError): self.test_obj['All'].get(tstart=3., tstop=3.) # tstart should be < tstop
+        np.testing.assert_allclose(np.array(self.test_obj['All'].get(node_ids=[3, 4], tstart=0.2, tstop=0.4).data[0]), [11.0, 11.1, 11.2, 11.3, 11.4, 11.5, 11.6, 11.7, 11.8, 11.9], 1e-6, 0)
```
tomdele: Can you do the same but for the SpikeReport ?

python/tests/test.py
```diff
@@ -307,12 +310,14 @@ def test_get_reports_from_population(self):
...
         self.assertEqual(keys, [(13, 30), (13, 30), (13, 31), (13, 31), (13, 32), (14, 32), (14, 33), (14, 33), (14, 34), (14, 34)])
 
-        self.assertEqual(len(sel.times), 3)  # Number of timestamp (0.8, 1.0 and 1.2)
-        sel = self.test_obj['All'].get(tstart=5., tstop=-1)  # tstart out of range
-        self.assertEqual(len(sel.ids), 0)
-        sel = self.test_obj['All'].get(tstart=3., tstop=3.)
-        self.assertEqual(len(sel.ids), 100)
+        self.assertEqual(len(self.test_obj['All'].get(node_ids=[]).data), 0)
+        self.assertEqual(len(self.test_obj['All'].get(node_ids=[]).times), 0)
+        self.assertEqual(len(self.test_obj['All'].get(node_ids=[]).ids), 0)
```
tomdele, line 315: Can you do the same but for the SpikeReport ?

alkino:
It works for me

tomdele:
The time index is not empty

tomdele:
actually I have this : 

```
In [1]: from libsonata import SpikeReader                                                                                                                                                                   
In [2]: a = SpikeReader("spikes.h5")                                                                                                                                                                        
In [3]: b= a["default"]                                                                                                                                                                                     
In [4]: b.get(node_ids=[])                                                                                                                                                                                  
Out[4]: [(2, 0.1), (0, 0.2), (1, 0.3), (2, 0.7), (0, 1.3)]
```
maybe I messed up ?

tomdele:
just add the same tests as the Frame version and then we will be sure !

python/tests/test.py
```diff
@@ -307,12 +312,14 @@ def test_get_reports_from_population(self):
...
 
+        self.assertEqual(len(sel.times), 3)  # Number of timestamp (0.8, 1.0 and 1.2)
+        with self.assertRaises(SonataError): self.test_obj['All'].get(tstart=5.)  # tstart out of range
+        self.test_obj['All'].get(tstart=3., tstop=3.) # tstart should be < tstop
```
WeinaJi, line 321: Last nitpick, please change the comment here since now it is allowed tstart = stop.

tomdele: good catch ! Actually you should have raised here no ?

WeinaJi: I requested Nico to allow ```tstart == stop``` so that users can retrieve the data for one single timestamp (In fact I am using that in the circuit-faker benchmark test). So it should not raise error here.

tomdele: ahhh I did not recall correctly and thought you asked for the tstart == stop to raise ... 
Ok !

alkino:
So I think we can merge it with the fix of @ferdonline ?

tomdele:
can you just add a simple test to check the double accessor for the data and the times ? so we can t regress on this ?

alkino:
it's done

tomdele:
For time not data. If you want to change the bindings one day for matrix we don t want to regress

alkino:
https://github.com/BlueBrain/libsonata/pull/96/commits/927f4c1bded987940e7f258eb72b329e469f20b0#diff-86df06a50f54cc01572d0a1fe4ca6f69R304

alkino:
Ok, let's go

tomdele:
I am good

PR #99, opened 2020-07-10
Expose Report data as Numpy

alkino:
## The problem
The `DataFrame` struct contains the Report data following a query. These can be long vectors, and default Pybind11 interface is a list. Converting such data to list, besides heavy in memory and processing, requires subsequent conversion to numpy or Pandas to be efficiently handled.

## This PR
Exposes the larger `.times` and `.data` vectors as Numpy arrays, without copies. 
Initial attempt was using the `asArray` method. However it steals the memory from the struct, which is not acceptable as subsequent requests for the vector would fail.
Second attempt was using memory-views. Although simple and functional, this approach suffered from the fact that memory views have a rudimentary API and therefore would require later conversion to numpy. Besides, when deleting the struct object the memory view would go invalid.
Third and clean option is to return a Numpy array with the custom handle. In case the handle is a python object, numpy will automatically adjust reference counting on that object, the owner of the data, in this case the struct. It is a perfect fit, however, to retrieve the python object from the current C++ object one requires internal API. A new helper function `managedMemoryArray` was created to return such arrays.

### DIFF
--- a/include/bbp/sonata/report_reader.h
+++ b/include/bbp/sonata/report_reader.h
@@ -23,7 +23,8 @@ struct SONATA_API DataFrame {
     std::vector<double> times;
     DataType ids;
     // data[times][ids]
-    std::vector<std::vector<float>> data;
+    size_t n_cols, n_rows;
+    std::vector<float> data;
 };
 
 using Spike = std::pair<NodeID, double>;
--- a/include/bbp/sonata/report_reader.h
+++ b/include/bbp/sonata/report_reader.h
@@ -22,8 +22,7 @@ struct SONATA_API DataFrame {
     using DataType = std::vector<KeyType>;
     std::vector<double> times;
     DataType ids;
-    // data[times][ids]
-    size_t n_cols, n_rows;
+    // data[times][ids], flattened. n_cols is ids.size()
     std::vector<float> data;
 };
 
--- a/python/bindings.cpp
+++ b/python/bindings.cpp
@@ -322,7 +322,9 @@ void bindReportReader(py::module& m, const std::string& prefix) {
                                                   dframe.n_cols,
                                                   dframe.n_rows);
                                })
-        .def_readonly("times", &DataFrame<KeyType>::times);
+        .def_property_readonly("times", [](DataFrame<KeyType>& dframe) {
+            return asArray(std::move(dframe.times));
+        });
     py::class_<typename ReportType::Population>(m,
                                                 (prefix + "ReportPopulation").c_str(),
                                                 "A population inside a ReportReader")
--- a/python/bindings.cpp
+++ b/python/bindings.cpp
@@ -51,10 +51,14 @@ py::array asArray(std::vector<std::string>&& values) {
     return py::array(py::dtype("object"), ptr->size(), ptr->data(), freeWhenDone(ptr));
 }
 
+
+// Return a 2D structure from a flat std::vector
 template <typename T>
-py::array asArray(std::vector<T>&& values, ssize_t n_cols, ssize_t n_rows) {
+py::array asArray(std::vector<T>&& values, ssize_t n_cols) {
+    assert(values.size() % n_cols == 0);
+    const ssize_t n_rows = values.size() / n_cols;
     auto ptr = new std::vector<T>(std::move(values));
-    return py::array({n_cols, n_rows}, ptr->data(), freeWhenDone(ptr));
+    return py::array({n_rows, n_cols}, ptr->data(), freeWhenDone(ptr));
 }
 
 
@@ -316,12 +320,9 @@ void bindReportReader(py::module& m, const std::string& prefix) {
                                    (prefix + "DataFrame").c_str(),
                                    "Something easily convertible to pandas dataframe")
         .def_readonly("ids", &DataFrame<KeyType>::ids)
-        .def_property_readonly("data",
-                               [](DataFrame<KeyType>& dframe) {
-                                   return asArray(std::move(dframe.data),
-                                                  dframe.n_cols,
-                                                  dframe.n_rows);
-                               })
+        .def_property_readonly("data", [](DataFrame<KeyType>& dframe) {
+            return asArray(std::move(dframe.data), dframe.ids.size());
+        })
         .def_property_readonly("times", [](DataFrame<KeyType>& dframe) {
             return asArray(std::move(dframe.times));
         });
--- a/python/bindings.cpp
+++ b/python/bindings.cpp
@@ -51,6 +51,12 @@ py::array asArray(std::vector<std::string>&& values) {
     return py::array(py::dtype("object"), ptr->size(), ptr->data(), freeWhenDone(ptr));
 }
 
+template <typename T>
+py::array asArray(std::vector<T>&& values, ssize_t n_cols, ssize_t n_rows) {
+    auto ptr = new std::vector<T>(std::move(values));
+    return py::array({n_cols, n_rows}, ptr->data(), freeWhenDone(ptr));
+}
+
 
 template <typename T>
 py::object getAttribute(const Population& obj,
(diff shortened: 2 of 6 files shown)


tomdele:
Actually if you can do that for ids and times it would be awesome.

tomdele:
I benched the perf using the pandas interface from bluepysnap : 

```
from bluepysnap import Simulation
import numpy as np
import timeit

if __name__=="__main__":
    sim = Simulation("simulation_config.json")
    r = sim.reports["bbprep0"]["node_population_1"]
    print(str(timeit.timeit(lambda: r.get(group=np.arange(900), t_start=0.0, t_stop=1000.0), number=1)) + " s")
```

Results :
List version from libsonata with no conversion to numpy array before the Dataframe constructor :
```
6.4 s
```
List version from libsonata with the user carefully changing the data to numpy before adding it to the DataFrame constructor:
```
0.4957452419985202 s
```
With the numpy array directly from libsonata:
```
0.27181885599929956 s
```

So basically if your consumer is not super careful with the current list outputted from libsonata and the pandas constructor used, one can have a factor 20 in perf

With the numpy array there is no risk of users poorly handling the outputs

tomdele:
@ferdonline I am a pure noob with numpy bindings ... Can you review this also?

python/bindings.cpp
```diff
@@ -51,6 +51,12 @@ py::array asArray(std::vector<std::string>&& values) {
     return py::array(py::dtype("object"), ptr->size(), ptr->data(), freeWhenDone(ptr));
 }
 
+template <typename T>
+py::array asArray(std::vector<T>&& values, ssize_t n_cols, ssize_t n_rows) {
+    auto ptr = new std::vector<T>(std::move(values));
+    return py::array({n_cols, n_rows}, ptr->data(), freeWhenDone(ptr));
```
ferdonline, line 57: I find this a bit unsafe, one could go out of bounds easily.
What about passing a single parameter, the one of the width?
```suggestion
py::array asArray(std::vector<T>&& values, ssize_t n_cols) {
    assert values.size() % n_cols == 0,
           "Data mismatch. Values size be a multiple of n_cols"
    const auto ssize_t n_rows = values.size() / n_cols;
    auto ptr = new std::vector<T>(std::move(values));
    return py::array({n_cols, n_rows}, ptr->data(), freeWhenDone(ptr));
```

python/bindings.cpp
```diff
@@ -310,8 +316,15 @@ void bindReportReader(py::module& m, const std::string& prefix) {
...
                                    "Something easily convertible to pandas dataframe")
         .def_readonly("ids", &DataFrame<KeyType>::ids)
-        .def_readonly("data", &DataFrame<KeyType>::data)
-        .def_readonly("times", &DataFrame<KeyType>::times);
+        .def_property_readonly("data",
+                               [](DataFrame<KeyType>& dframe) {
+                                   return asArray(std::move(dframe.data),
+                                                  dframe.n_cols,
+                                                  dframe.n_rows);
+                               })
```
ferdonline, line 324: just consistency with previous style
```suggestion
        .def_property_readonly("data", [](DataFrame<KeyType>& dframe) {
            return asArray(std::move(dframe.data), dframe.n_cols, dframe.n_rows);
        })
```

include/bbp/sonata/report_reader.h
```diff
@@ -23,7 +23,8 @@ struct SONATA_API DataFrame {
     std::vector<double> times;
     DataType ids;
     // data[times][ids]
-    std::vector<std::vector<float>> data;
+    size_t n_cols, n_rows;
```
ferdonline, line 26: `n_rows` could be extracted from the data.size() (/n_cols) to avoid inconsistencies

include/bbp/sonata/report_reader.h
```diff
@@ -23,7 +23,8 @@ struct SONATA_API DataFrame {
...
+    std::vector<float> data;
```
ferdonline, line 27: Do we have to live without other libraries? What about `boost::numeric::ublas::matrix`? It offers a simple API and ensures memory contiguity.

python/bindings.cpp
ferdonline: Btw I think I still would prefer using a dedicated type for matrix, other than this `n_cols` workaround

include/bbp/sonata/report_reader.h
matz-e: I think this is sufficient, don't see the point in adding the heavy-handed Boost dependency. But as you said, one of rows/cols could be inferred all the time. As this is contained internally, I don't see it as a big issue, though.

pramodk: (If there is no existing boost dependency, I would definitely avoid it)

src/report_reader.cpp
```diff
@@ -369,24 +397,12 @@ DataFrame<T> ReportReader<T>::Population::get(const nonstd::optional<Selection>&
             .read(data);
         int timer_index = 0;
         for (const std::vector<float>& datum : data) {
-            for (float d : datum) {
-                data_frame.data[timer_index].push_back(d);
-            }
+            std::copy(datum.data(),
+                      datum.data() + datum.size(),
+                      &data_frame.data[timer_index * data_frame.n_rows + ids_index]);
```
ferdonline, line 402: I believe there was a bug here. The flattened structure changes row by doing `index*n_cols`, not `index*n_rows`

ferdonline: Actually there is a general naming swap rows-columns... Improving this

python/bindings.cpp
```diff
@@ -316,12 +320,9 @@ void bindReportReader(py::module& m, const std::string& prefix) {
                                    (prefix + "DataFrame").c_str(),
                                    "Something easily convertible to pandas dataframe")
         .def_readonly("ids", &DataFrame<KeyType>::ids)
-        .def_property_readonly("data",
-                               [](DataFrame<KeyType>& dframe) {
-                                   return asArray(std::move(dframe.data),
-                                                  dframe.n_cols,
-                                                  dframe.n_rows);
-                               })
+        .def_property_readonly("data", [](DataFrame<KeyType>& dframe) {
```
matz-e, line 323: Stupid question… is the `data` property cached? Otherwise, due to the `std::move`, this would be a one-shot operation.

python/bindings.cpp
```diff
@@ -51,10 +51,14 @@ py::array asArray(std::vector<std::string>&& values) {
     return py::array(py::dtype("object"), ptr->size(), ptr->data(), freeWhenDone(ptr));
 }
 
+
+// Return a 2D structure from a flat std::vector
 template <typename T>
-py::array asArray(std::vector<T>&& values, ssize_t n_cols, ssize_t n_rows) {
+py::array asArray(std::vector<T>&& values, ssize_t n_cols) {
+    assert(values.size() % n_cols == 0);
```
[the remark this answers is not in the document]
ferdonline: I am not sure we do that, but the purpose here is more for self-documentation and debugging other than enforcement at run-time.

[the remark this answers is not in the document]
matz-e: OK

python/bindings.cpp
ferdonline: that's a very good point. `asArray` is typically used right after we extracted the data from some place, to avoid making copies of temp data. Indeed it seems the source struct is not temp here... so the user can't call it twice. 
Maybe the data is too large to make copies and we should attempt to share/cache the array?

python/bindings.cpp
```diff
@@ -52,6 +52,26 @@ py::array asArray(std::vector<std::string>&& values) {
 }
 
 
+// Return a 2D structure from a flat std::vector
+template <typename T>
+py::array asArray(std::vector<T>&& values, ssize_t n_cols) {
```
matz-e, line 57: Couldn't this be deleted then?

ferdonline: True, not needed for the moment. You think it's not worth to keep, just in case for the future?

matz-e: I guess that's OK, although I feel like it's a little dangerous, still.

ferdonline:
@tomdele could you run your benchmarks again please?

tomdele:
Yes sure

tomdele:
I have the same results as before : 

0.2783471200000349 s

tomdele:
Is it expected ?

ferdonline:
yes! 🎉  thanks!
Now it should be fine as well if you access the property twice.

tomdele:
I am glad we asked you to check :smile:

tomdele:
I checked the result is the same with multiple calls. 
Can you add a test with multiple calls ?

PR #102, opened 2020-08-25
Fix bug when reading empty spike reports

jorblancoa:


alkino (approved):

src/report_reader.cpp
```diff
@@ -172,8 +172,8 @@ SpikeReader::Population::Population(const std::string& filename,
     }
 
     if (sorting_ == Sorting::by_time) {
-        tstart_ = timestamps.front();
-        tstop_ = timestamps.back();
+        tstart_ = timestamps.empty() ? 0 : timestamps.front();
+        tstop_ = timestamps.empty() ? 0 : timestamps.back();
     } else {
         tstart_ = *min_element(timestamps.cbegin(), timestamps.cend());
```
mgeplf, line 178: could you add a test?

does this apply to the other `Sorting` methods?

mgeplf (changes requested):

src/report_reader.cpp
jorblancoa: You are right, I added a test and fix for all sorting methods.

mgeplf (approved):
Thanks, looks good.

PR #104, opened 2020-10-09
Improves IO performance for the report readers

sergiorg-hpc:
### **Changes**

**IO:**
  * Loop through rows (timesteps) instead of columns (node_ids).
      - Changes from tsteps * node_ids reads, to tsteps reads (i.e.,
        constant time in terms of IO, regardless of node_ids reqs.).
  * Read all node_ids in a buffer every timestep, between min and max
    offsets of node_ids requested.

**Memory:**
  * Allocate buffer before passing to HighFive to avoid extra and
    unnecessary memory allocations before reading.
  * Change data structure of node_ids/offsets from vector to map
    for faster search (from n^2 to nlogn).
  * In soma reports, assign values directly to return buffer instead
    of std::copy/memcpy.

**New features and others:**
  * Add support for strided reads, allowing to reduce the amount of
    timesteps (e.g., 1 by default, 2 for every 2 timesteps, etc.).
  * Eliminate duplicated code, avoid calling HDF5 metadata in every
    iteration, updated unit tests, and other minor changes.

### **Important note**
Performance evaluations and other related discussions can be obtained here:
https://bbpteam.epfl.ch/project/issues/browse/REP-60
https://bbpteam.epfl.ch/project/issues/browse/BLPY-217


src/report_reader.cpp
```diff
@@ -349,22 +349,25 @@ DataFrame<T> ReportReader<T>::Population::get(const nonstd::optional<Selection>&
...
         }
-
-        std::vector<ElementID> element_ids;
-        pop_group_.getGroup("mapping")
-            .getDataSet("element_ids")
-            .select({it->second.first}, {it->second.second - it->second.first})
-            .read(element_ids);
+        if (it->second.first < min) {
+            min = it->second.first;
+        }
```
alkino, line 363: oneline?
min = std::min(it->second.first, min);

src/report_reader.cpp
```diff
@@ -349,22 +349,25 @@ DataFrame<T> ReportReader<T>::Population::get(const nonstd::optional<Selection>&
...
+        if (it->second.second > max) {
+            max = it->second.second;
+        }
```
alkino, line 366: idem with std::max

alkino (dismissed):

alkino:
Really nice improvment

alkino:
Really nice way to improve things

src/report_reader.cpp
```diff
@@ -374,43 +374,34 @@ DataFrame<T> ReportReader<T>::Population::get(const nonstd::optional<Selection>&
...
-
-        int timer_index = 0;
-
-        for (const std::vector<float>& datum : data) {
-            std::copy(datum.data(),
-                      datum.data() + datum.size(),
-                      &data_frame.data[timer_index * n_ids + ids_index]);
-            ++timer_index;
+    std::vector<float> buffer(max - min);
+    auto dataset = pop_group_.getDataSet("data");
```
matz-e, line 382: Nice!

matz-e (approved):
Great! Some nitpicks about the docs… if that's intentional, please ignore.

include/bbp/sonata/report_reader.h
```diff
@@ -123,19 +123,22 @@ class SONATA_API ReportReader
...
+                               const nonstd::optional<size_t>& tstride = nonstd::nullopt) const;
 
       private:
         Population(const H5::File& file, const std::string& populationName);
-        std::pair<size_t, size_t> getIndex(const nonstd::optional<double>& tstart, const nonstd::optional<double>& tstop) const;
+        std::pair<size_t, size_t> getIndex(const nonstd::optional<double>& tstart,
+                                           const nonstd::optional<double>& tstop) const;
 
-        std::vector<std::pair<NodeID, std::pair<uint64_t, uint64_t>>> nodes_pointers_;
+        std::map<NodeID, std::pair<uint64_t, uint64_t>> nodes_pointers_;
```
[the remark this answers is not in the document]
alkino: The idea is to use the black red tree to do fast search inside NodeID.
NodeID can be 2 millions and we need to select maybe 10,000 ids.
They are not sorted.

`nodes_pointers_` is a sorted cache

src/report_reader.cpp
```diff
@@ -374,43 +374,34 @@ DataFrame<T> ReportReader<T>::Population::get(const nonstd::optional<Selection>&
...
+    for (size_t timer_index = index_start; timer_index <= index_stop; timer_index += stride) {
+        dataset.select({timer_index, min}, {1, max - min}).read(buffer.data());
```
GianlucaFicarelli, line 384: Nice solution. This should work well for contiguous data, or horizontal chunks, while I don't expect good performance with vertical chunks. If this is the case (it could be tested), maybe it's possible to add a comment here or in the documentation?

sergiorg-hpc: The solution is meant for the current column-based GID access, and it is supposed to avoid any changes in the HDF5 files (i.e., no chunking or any other external trick). The reader is meant to be as generic as possible for any kind of application, including for visualization purposes where vertical chunking would affect their performance.

We will discuss these matters in detail during our presentation, but thanks for the insight!

src/report_reader.cpp
```diff
@@ -349,22 +349,22 @@ DataFrame<T> ReportReader<T>::Population::get(const nonstd::optional<Selection>&
         node_ids = selection->flatten();
```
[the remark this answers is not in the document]
alkino: why?

include/bbp/sonata/report_reader.h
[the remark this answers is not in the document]
jorblancoa: Exactly!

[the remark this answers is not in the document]
GianlucaFicarelli: I'm wondering: shouldn't the performance of `unordered_map` (unordered, O(1) access time complexity) and sorting the results only once and only when all the nodes are selected (i.e. `selection` is not specified when calling `get`) be better than using `map` (ordered, O(log N) access time complexity)?
Anyway the memory usage would be higher.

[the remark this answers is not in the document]
sergiorg-hpc: The point is that, while the unordered map can technically provide certain advantages, a higher amount of GIDs can also increase the chances for conflicts in the buckets of the hash table. This means that it could potentially be a sequential O(N) search of certain elements.

Nonetheless, we tried to use first an unordered map and the performance was about the same as a conventional map. Thus, we kept the conventional map to guarantee the same access time and lower the memory footprint. In addition, right now the search for the GIDs amounts for a very little piece of the overall execution time, even for 100K elements.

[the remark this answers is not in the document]
matz-e: OK, was curious… I guess in the end IO is probably going to overwhelm any performance tuning here, so KISS is better. Thanks for the follow-up!

src/report_reader.cpp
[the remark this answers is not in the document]
GianlucaFicarelli: Using `reserve` we will avoid the memory re-allocation for the vector `node_ids` when it's populated.
Since we know the size of the final vector, we could do this before the `transform` loop.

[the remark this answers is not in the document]
alkino: But here, the function produce a vector of the right size and as this is a prvalue, it will be moved.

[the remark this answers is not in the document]
alkino: Maybe I don't understand well because your link seems broken.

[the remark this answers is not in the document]
GianlucaFicarelli: The link is to the (current) line 340 `std::transform(nodes_pointers_.begin(),`.
Sorry but github does not allow me to comment on that line because it was not modified in this PR.

include/bbp/sonata/report_reader.h
```diff
@@ -123,19 +123,22 @@ class SONATA_API ReportReader
 
         /**
          * \param node_ids limit the report to the given selection.
-         * \param tstart return spikes occurring on or after tstart. tstart=nonstd::nullopt
-         * indicates no limit. \param tstop return spikes occurring on or before tstop.
-         * tstop=nonstd::nullopt indicates no limit.
+         * \param tstart return voltages occurring on or after tstart. tstart=nonstd::nullopt
+         * indicates no limit. \param tstop return voltages occurring on or before tstop.
```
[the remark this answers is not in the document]
jorblancoa: I just kept the same format... 
Before tstart and tstop were without line break so I did the same thing.

src/report_reader.cpp
pramodk: > This should work well for contiguous data, or horizontal chunks, while I don't expect good performance with vertical chunks.

@GianlucaFicarelli : In [this plot](https://bbpteam.epfl.ch/project/issues/browse/BLPY-217?focusedCommentId=127949&page=com.atlassian.jira.plugin.system.issuetabpanels%3Acomment-tabpanel#comment-127949), selecting single gid with all timeframes is vertical selection and performance is acceptable. Does that answer your question?

Obviously there are other improvements we would like to do but but those will come in separate PR.

GianlucaFicarelli: @pramodk I was meaning something different, I'll try to explain better: the same data can be stored in a h5 file contiguously, or in chunks of different shape and size.
1. If data is contiguous (and not transposed), then it's stored in C order (row-major) as (t0,gid0), (t0,gid1), (t0,gid2)..., and performance to read by rows with the solution in this PR is good.
2. If instead data is stored in vertical chunks (for example 1 gid and all its time frames in one chunk), then chunk 0 contains (t0,gid0), (t1,gid0), (t2,gid0)... In this case the solution of this PR has not good performance (I did some tests to confirm this).
3. Finally, with horizontal chunks (for example 1 chunk contains 1 time frame, and all the gids), the performance with the solution in this PR looks a bit worse than the contiguous case (i.e. 300 seconds vs 80 for 10K samples in my tests).
Anyway I think that the solution of this PR is good because it does not require any change to the contiguous h5 format that we are using.
I just think that it can be mentioned in the documentation that h5 chunking can affect performances.

alkino: data are normalized: https://github.com/AllenInstitute/sonata/blob/master/docs/SONATA_DEVELOPER_GUIDE.md#frame-oriented-node-element-recordings

So they can be chunked, but not vertically oriented. This library only support this.

We have to find why this is worse if it is chunked.

pramodk: Ok, I now understand your point. 

> If instead data is stored in vertical chunks (for example 1 gid and all its time frames in one chunk), then chunk 0 contains (t0,gid0), (t1,gid0), (t2,gid0)... In this case the solution of this PR has not good performance (I did some tests to confirm this).

If the writer parts will be changed in the future, for sure we will need to reconsider optimisations. For example, if we chunk by column, the data for all time steps for specific gid will be read in one go and there won't be need to do filtering or finding min/max.

src/report_reader.cpp
```diff
@@ -315,18 +315,18 @@ std::pair<size_t, size_t> ReportReader<T>::Population::getIndex(
...
-                                              const nonstd::optional<double>& tstop) const {
+                                              const nonstd::optional<double>& tstop,
+                                              const nonstd::optional<size_t>& tstride) const {
     DataFrame<T> data_frame;
-
     size_t index_start = 0;
     size_t index_stop = 0;
     std::tie(index_start, index_stop) = getIndex(tstart, tstop);
-
+    const size_t stride = tstride.value_or(1) > 0 ? tstride.value_or(1) : 1;
```
mgeplf, line 324:
```suggestion
    const size_t stride = tstride.value_or(1);
    if(stride == 0) { throw ... }
```

src/report_reader.cpp
```diff
@@ -349,22 +349,22 @@ DataFrame<T> ReportReader<T>::Population::get(const nonstd::optional<Selection>&
         node_ids = selection->flatten();
     }
 
+    std::vector<std::pair<uint64_t, uint64_t>> positions;
+    uint64_t min = UINT64_MAX;
+    uint64_t max = 0;
```
mgeplf, line 354: The (nice) commit msg and presentation mentions why this min/max calc is happening, but perhaps add a small comment?

src/report_reader.cpp
mgeplf, line 382: what happens if the `data` dataset isn't float?

src/report_reader.cpp
```diff
@@ -374,43 +374,34 @@ DataFrame<T> ReportReader<T>::Population::get(const nonstd::optional<Selection>&
...
+        for (const auto& position : positions) {
+            uint64_t elements_per_gid = position.second - position.first;
+            uint64_t gid_start = position.first - min;
+            uint64_t gid_end = position.second - min;
+
+            // Soma report
+            if (elements_per_gid == 1) {
+                data_ptr[offset] = buffer[gid_start];
+            } else {  // Elements report
+                std::memcpy(&data_ptr[offset],
```
mgeplf, line 398: do we have to use `memcpy`?  The potential for errors is quite high, especially w/ the `sizeof(float) * (gid_end - gid_start)` idiom.  `positions` needs to be strongly validated against the len of dataframe.data, otherwise we're in UB land.

mgeplf (commented):
Very nice addition.

src/report_reader.cpp
```diff
@@ -349,22 +349,22 @@ DataFrame<T> ReportReader<T>::Population::get(const nonstd::optional<Selection>&
...
+    uint64_t max = 0;
+    auto dataset_elem_ids = pop_group_.getGroup("mapping").getDataSet("element_ids");
     for (const auto& node_id : node_ids) {
-        const auto it = std::find_if(
-            nodes_pointers_.begin(),
-            nodes_pointers_.end(),
-            [&node_id](const std::pair<NodeID, std::pair<NodeID, uint64_t>>& node_pointer) {
-                return node_pointer.first == node_id;
-            });
+        const auto it = nodes_pointers_.find(node_id);
```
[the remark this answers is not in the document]
alkino: yes, if I remember correctly this is user input and they can ask for missing node_id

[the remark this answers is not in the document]
mgeplf: yes, good point, thanks.

src/report_reader.cpp
alkino: Extract from https://github.com/AllenInstitute/sonata/blob/master/docs/SONATA_DEVELOPER_GUIDE.md#frame-oriented-node-element-recordings

> /report/<population_name>/data (dtype:float, shape: N_time x N_values): Writers are encouraged to use chunking for efficient read access. Attributes:

src/report_reader.cpp
```diff
@@ -349,22 +349,22 @@ DataFrame<T> ReportReader<T>::Population::get(const nonstd::optional<Selection>&
         node_ids = selection->flatten();
     }
 
+    std::vector<std::pair<uint64_t, uint64_t>> positions;
```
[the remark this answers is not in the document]
sergiorg-hpc: This generally helps with readability, although in this particular case it can be confusing. Do you have a name in mind for the alias?

For instance, if we change it here, we must change it in the `std::map<<NodeID, std::pair<uint64_t, uint64_t>> nodes_pointers_`.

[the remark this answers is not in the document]
mgeplf: Yeah, I meant for it to be everywhere.  It can be internal to the `SpikeReader`, and something similar to the alias in population.h: `using Range = std::pair<uint64_t, uint64_t>`.  Was just a thought, up to you.

src/report_reader.cpp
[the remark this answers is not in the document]
sergiorg-hpc: Thanks @GianlucaFicarelli, we just added your change!

src/report_reader.cpp
alkino: #105 has been open about lower performances with chunked files.

src/report_reader.cpp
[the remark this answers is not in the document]
sergiorg-hpc: Thanks @mgeplf, you're right. We added tests to coverage that the output is given in order. We tested to change the structure to `std::unordered_map` and the tests fail as expected.

src/report_reader.cpp
sergiorg-hpc: We added a comment in the source code, just to ensure future changes consider your suggestion, @GianlucaFicarelli . Thanks for the input!

src/report_reader.cpp
```diff
@@ -374,43 +374,34 @@ DataFrame<T> ReportReader<T>::Population::get(const nonstd::optional<Selection>&
...
+    for (size_t timer_index = index_start; timer_index <= index_stop; timer_index += stride) {
+        dataset.select({timer_index, min}, {1, max - min}).read(buffer.data());
+
+        off_t offset = 0;
+        off_t data_offset = (timer_index - index_start) / stride;
+        auto data_ptr = &data_frame.data[data_offset * n_ids];
+        for (const auto& position : positions) {
+            uint64_t elements_per_gid = position.second - position.first;
+            uint64_t gid_start = position.first - min;
+            uint64_t gid_end = position.second - min;
```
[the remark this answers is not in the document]
sergiorg-hpc: Agreed!

src/report_reader.cpp
sergiorg-hpc: Done! Thanks for the suggestion.

src/report_reader.cpp
sergiorg-hpc: We changed to `memcpy` because for single elements the overhead of `std::copy` was considerable. But as now we only use it for element reports, we changed it back as you suggested, agreed.

P.S.: Just to clarify, we did a small test application that copies 200-300 `float` elements at once, many times, to replicate the same behaviour of the reader implementation. The difference in performance between both versions was negligible.

sergiorg-hpc:
@jorblancoa and I went through all the comments and suggestions, and integrated everything (specially Jorge, thanks for the effort).

If no one has any additional comments, we are ready to close the PR and merge the changes!

alkino (dismissed):

src/report_reader.cpp
mgeplf: Can check that this is true?  If someone provides an hdf file that doesn't meet the spec, it could cause us to read outside the bounds, no?

include/bbp/sonata/report_reader.h
```diff
@@ -138,7 +138,7 @@ class SONATA_API ReportReader
         std::pair<size_t, size_t> getIndex(const nonstd::optional<double>& tstart,
                                            const nonstd::optional<double>& tstop) const;
 
-        std::map<NodeID, std::pair<uint64_t, uint64_t>> nodes_pointers_;
+        std::map<NodeID, Selection::Range> nodes_pointers_;
```
mgeplf, line 141: I would use a local type, not the `Selection::Range`, as its behavior might change.  Semantically they operate the same now, but that's not guaranteed to be the case; this isn't really a `Selection`

python/tests/test.py
```diff
@@ -275,15 +275,22 @@ def test_get_reports_from_population(self):
...
         self.assertEqual(len(self.test_obj['All'].get().ids), 20)  # Number of nodes
         self.assertEqual(len(self.test_obj['All'].get().times), 10)  # number of times
         self.assertEqual(len(self.test_obj['All'].get().data), 10)  # should be the same
+
         sel = self.test_obj['All'].get(node_ids=[13, 14], tstart=0.8, tstop=1.0)
         self.assertEqual(len(sel.times), 2)  # Number of timestamp (0.8 and 0.9)
         self.assertEqual(list(sel.ids), [13, 14])
         np.testing.assert_allclose(sel.data, [[13.8, 14.8], [13.9, 14.9]])
 
+        sel_all = self.test_obj['All'].get()
```
mgeplf, line 288: thanks

include/bbp/sonata/report_reader.h
sergiorg-hpc: Agreed, it was just to avoid duplicating code, but we can change it.

src/report_reader.cpp
jorblancoa: The generated report from the simulation will be always float. However it is true that if someone is trying to read a pseudo-compatible report it might be an issue.

An idea would be to check the datatype and if it is not float to throw an exception...

mgeplf: > An idea would be to check the datatype and if it is not float to throw an exception...

That's the safest bet, imo.

mgeplf: Nice! Thanks.

mgeplf (approved):
LGTM, nice work!

PR #106, opened 2020-10-21
allow tests to be run from any location

mgeplf:


matz-e (approved):

WeinaJi (approved):
Yes, it looks nicer in this way. Thanks for improving it.

PR #107, opened 2020-10-29
Implemented `node_sets` functionality

mgeplf:
* load JSON defined node sets
* materialize `Selection`s against `Population`s


mgeplf:
FYI I'm versioning in `https://github.com/nlohmann/json/` b/c the repo is large (~250MB for full checkout).  That causes clang-format to fail.

mgeplf:
@tomdele: how hard would it be to use your `bluepysnap` node_set implementation to test against?  Could you see if the results are the same?

src/node_sets.cpp
```diff
@@ -0,0 +1,317 @@
...
+{
+  public:
+    NodeSetBasicRule(std::string attribute, std::vector<T>&& values)
+        : attribute_(std::move(attribute))
+        , values_(values) {}
+
+    Selection materialize(const NodeSets& /* unused */, const NodePopulation& np) const final {
+        Selection ret{{}};
+        for (const auto& v : values_) {
+            ret = ret | np.matchAttributeValues(attribute_, v);
```
tomdele, line 41: maybe add a comment inside the function to remind the dev the float comparisons are not allowed.

src/node_sets.cpp
```diff
@@ -0,0 +1,317 @@
...
+
+class NodeSetBasicNodeIds: public NodeSetRule
+{
+  public:
+    explicit NodeSetBasicNodeIds(Selection::Values&& values)
+        : values_(values) {}
+
+    Selection materialize(const NodeSets& /* unused */,
+                          const NodePopulation& /* unused */) const final {
+        return Selection::fromValues(values_.begin(), values_.end());
```
tomdele, line 86: To me inside the nodesets all "[]" == "or". And I think it makes sense from a user point of view.
So to me,  for consistency, the node_id values should act like a "or" also and so you should only return the Selection of nodes actually inside the node population.

src/node_sets.cpp
```diff
@@ -0,0 +1,317 @@
...
+
+  private:
+    Selection::Values values_;
+};
+
+using CompoundTargets = std::vector<std::string>;
+class NodeSetCompoundRule: public NodeSetRule
+{
+  public:
+    NodeSetCompoundRule(std::string attribute, const CompoundTargets& targets)
```
tomdele, line 101: you need a name also for the CompoundRule otherwise you cannot have a correct toJson()

src/node_sets.cpp
```diff
@@ -0,0 +1,317 @@
...
+    Selection materialize(const NodeSets& ns, const NodePopulation& np) const final {
+        Selection ret{{}};
+        for (const auto& target : targets_) {
+            ret = ret | ns.materialize(target, np);
+        }
+        return ret;
+    }
+
+    std::string toJSON() const final {
+        return toString("node_ids", targets_);
```
tomdele, line 114: should be toString(name_, targets_);

src/node_sets.cpp
```diff
@@ -0,0 +1,317 @@
...
+    }
+
+  private:
+    std::string attribute_;
+    CompoundTargets targets_;
+};
+
+int64_t get_integer_or_throw(const json& el) {
+    auto v = el.get<double>();
+    if (std::floor(v) != v) {
```
tomdele, line 124: std::numeric_limits<double>::epsilon ?

src/node_sets.cpp
```diff
@@ -0,0 +1,317 @@
...
+        if (el.value().is_number()) {
+            if (attribute == "population") {
+                throw SonataError("'population' must be a string");
+            }
+
+            int64_t v = get_integer_or_throw(el.value());
+            if (attribute == "node_id") {
+                if (v < 0) {
+                    throw SonataError("'node_id' must be positive");
+                }
```
tomdele, line 144: this will depends if you are using the "or" for the "node_id" key like I do in snap

src/node_sets.cpp
```diff
@@ -0,0 +1,317 @@
...
+
+                ret.emplace_back(new NodeSetBasicNodeIds({static_cast<Selection::Value>(v)}));
+            } else {
+                ret.emplace_back(new NodeSetBasicRule<int64_t>(attribute, {v}));
```
tomdele, line 148: is the <int64_t> ok if the parameter inside the is let say a int32 or uint32 ?

src/node_sets.cpp
```diff
@@ -0,0 +1,317 @@
...
+            } else {
+                ret.emplace_back(
+                    new NodeSetBasicRule<std::string>(attribute, {el.value().get<std::string>()}));
+            }
+        } else if (el.value().is_array()) {
+            const auto array = el.value();
+
+            if (attribute == "population" && (array.size() != 1 || !array[0].is_string())) {
+                throw SonataError("'population' must be a single string");
+            }
```
tomdele, line 166: see my comment above #L58

src/node_sets.cpp
```diff
@@ -0,0 +1,317 @@
...
+
+            if (array[0].is_number()) {
+                std::vector<int64_t> values;
+                for (auto& inner_el : array.items()) {
+                    values.emplace_back(get_integer_or_throw(inner_el.value()));
```
tomdele, line 171: for int is there a benefit in using emplace_back instead of push_back ?

src/node_sets.cpp
```diff
@@ -0,0 +1,317 @@
...
+            } else {
+                throw SonataError("Unknown array type");
+            }
+        }
+    }
+    return ret;
+}
+
+void parse_basic(const json& j, std::map<std::string, NodeSetRules>& node_sets) {
+    for (auto& el : j.items()) {
```
tomdele, line 209: auto const& el ?

src/node_sets.cpp
```diff
@@ -0,0 +1,317 @@
...
+}
+
+void parse_compound(const json& j, std::map<std::string, NodeSetRules>& node_sets) {
+    std::map<std::string, CompoundTargets> compound_rules;
+    for (auto& el : j.items()) {
+        if (el.value().is_array()) {
+            CompoundTargets targets;
+            for (const auto& name : el.value()) {
+                if (!name.is_string()) {
+                    throw SonataError("All compound node types must be strings");
```
tomdele, line 241: "All compound elements must be strings" ?

tomdele:
> @tomdele: how hard would it be to use your bluepysnap node_set implementation to test against? Could you see if the results are the same?

Applied to a population ?

src/node_sets.cpp
```diff
@@ -0,0 +1,317 @@
...
+        throw SonataError("Top level node_set must be an object");
+    }
+
+    // Need to two pass parsing the json so that compound lookup can rely
+    // on all the basic rules existing
+    parse_basic(j, node_sets_);
+    parse_compound(j, node_sets_);
+}
+
+Selection NodeSets::materialize(const std::string& name, const NodePopulation& population) const {
```
tomdele, line 272: maybe "apply" is a better name, no ?

tomdele:
Not so hard imao :

We can create a small circuit config : 
```
{
  "manifest": {
    "$BASE_DIR": ".",
  },
  "node_sets_file": "$BASE_DIR/node_sets.json",
  "networks": {
    "nodes": [
      {
        "nodes_file": "$BASE_DIR/nodes.h5",
        "node_types_file": ""
      }
    ],
    "edges": [
    ]
  }
}
```
And try to : 
```
import json 
from bluepysnap import Circuit

# with the node_sets.json : 
#{ 
#"NodeSet0": { "attr-Y": 21 },
#}

circuit = Circuit("config.json")
population = NodeStorage(os.path.join(BASE_DIR, 'nodes.h5')).open_population('nodes-A')

node_sets = json.load(open("node_sets.json"))

for name_, value in node_sets.items():
    sel = NodeSets(json.dumps(value)).materialize(name_, population)
    assert circuit.nodes[name_].ids("NodeSet0") == sel
```
I guess

src/node_sets.cpp
tomdele: hum no sorry maybe just the targets separated with a coma

src/node_sets.cpp
tomdele: Ok sorry got it finally ... 

Actually attribute is the name of the compound rule, isn't it ?
So I guess you should rename attribute into name or something and  toString to  toString(name_, targets_);

tomdele: Ok sorry got it finally ... 

Actually attribute is the name of the compound rule, isn't it ?
So I guess you should rename attribute into name or something and  toString to  toString(name_, targets_);

src/node_sets.cpp
```diff
@@ -0,0 +1,317 @@
...
+
+  private:
+    std::string attribute_;
+    std::vector<T> values_;
+};
+
+class NodeSetBasicPopulation: public NodeSetRule
+{
+  public:
+    explicit NodeSetBasicPopulation(std::string name)
```
tomdele, line 58: should be names actually. 
To me there is no reason why you cannot have a list of populations with this : 
``` 
{
"pop1" : {"population" : "population1"},
"pop12" : {"population" : ["population1", "population2"]},
}
```
Allowing this would also the possibility to : 
``` 
{
"pop1" : {"population" : ["population1"]},
}
```
without raising

src/node_sets.cpp
mgeplf: I added it to the NodeSets docstring

src/node_sets.cpp
mgeplf: Yup, makes sense.

src/node_sets.cpp
mgeplf: sure

src/node_sets.cpp
mgeplf: Why? I want to make sure that it's an integer, not that it's close to an integer.

src/node_sets.cpp
mgeplf: Not sure what you mean; shouldn't `node_id` be positve?

src/node_sets.cpp
mgeplf: should be, it's a widening, which is safe, afaik.

src/node_sets.cpp
mgeplf: I think that comment got lost?

src/node_sets.cpp
mgeplf: Not in this case; but I'm using `emplace_back` elsewhere, and there's no cost, afaik.

src/node_sets.cpp
mgeplf: done

src/node_sets.cpp
mgeplf: sure

src/node_sets.cpp
tomdele: ah ok when I first read the function I though you wanted something close to an integer. 
Then don t you need abs() if the integer is a negative number ?

src/node_sets.cpp
tomdele: If the 'node_id' : [] act like a "or" then you could have (imao) [-5, 5, 6, -3] --> [5, 6] without raising. 
So yes node_id should be positive so you will never select the negative ids but I am not sure you should raise

src/node_sets.cpp
tomdele: but not for the uint64, no ?

src/node_sets.cpp
tomdele: This is the comment that have been killed during the UI madness ... 
It is part of a second round of comments. Put it here again : 

To me there is no reason why you cannot have a list of populations with this :
```
{
"pop1" : {"population" : "population1"},
"pop12" : {"population" : ["population1", "population2"]},
}
```
Allowing this would also allow to :
```
{
"pop1" : {"population" : ["population1"]},
}
```
without raising.

src/node_sets.cpp
tomdele: ok it was for my personal knowledge thx !

mgeplf:
@tomdele > Not so hard imao :
Cool, can you try it then?

src/node_sets.cpp
mgeplf: I'm agnostic, we'll see what others say.

src/node_sets.cpp
mgeplf: I think raising is appropriate: if someone puts a negative number, they should be warned, b/c it's probably a mistake.

src/node_sets.cpp
mgeplf: Negative numbers are allowed; it's not just `node_id` that can be tested against.

src/node_sets.cpp
mgeplf: I thought you were just asking about `int32`; yes, it would fail for `uint64` greater than 4294967296;  I'm not sure what the right behavior is in that case.

src/node_sets.cpp
mgeplf: ok, that's supported now.

src/node_sets.cpp
tomdele: You are using this here :

https://github.com/BlueBrain/libsonata/pull/107/files#diff-2eb0c657a8cf02cb85ca88ac45b71f6a3746c46ab114981bea14e605449a3325R138
So the NodeSetRules with an int attribute will also go inside this function.

So my question was : 
is `std::floor(v)  != v)` always safe for a negative number ? 
That is, is it possible `el.get<double> ` returns for some reason a number such as : -9.0000000000000000000000000001 and then the floor will be -10 instead of -9 ? 

Shouldn't we use : 
```
std::floor(std::abs(v))  != v
```
for safety ?

src/node_sets.cpp
tomdele: ok fair enough. This is just a behavior we should share among tools

tomdele:
> Cool, can you try it then?

sure. I will let you know when this is done

src/node_sets.cpp
tomdele: Yeah did not choose my example correctly ...

I think we can live with the uint64 pb for the moment. 
Just leave a comment so if one day this appears to be a problem we know where to look at

src/node_sets.cpp
mgeplf: > el.get<double> returns for some reason a number such as : -9.0000000000000000000000000001

All integers should be representable by float/doubles up to (about) 2**sizeof_mantissa, so `el.get<double>` shouldn't returning anything that requires `floor`ing, unless they aren't integers, regardless of sign.

tomdele: yeah indeed. 
All good then

tomdele:
Ok I used:

```
import json

from libsonata import NodeStorage, NodeSets
from bluepysnap import Circuit
import numpy as np

if __name__ == "__main__":

    circuit = Circuit("circuit_config.json")
    population = NodeStorage('nodes.h5').open_population('hippocampus_neurons')

    node_sets = json.load(open("node_sets.json"))
    sel = NodeSets(json.dumps(node_sets))

    for name_, value in node_sets.items():
        ids_libsonata = sel.materialize(name_, population).flatten()
        ids_bluepysnap = circuit.nodes[population.name].ids(name_)
        if len(ids_bluepysnap) == 0:
            print(f"no ids {name_}")
        diff_1 = np.setdiff1d(ids_bluepysnap, ids_libsonata)
        diff_2 = np.setdiff1d(ids_libsonata, ids_bluepysnap)
        assert len(diff_1) == len(diff_2) == 0
```
Using the MOOC node file (I added a int property to the file tho) and a custom node_set file the results are identical. 

You can check the nodesets + file + code I used here : 
/gpfs/bbp.cscs.ch/project/proj30/home/tdelemon/nodesets

python/tests/test.py
```diff
@@ -331,11 +336,109 @@ def test_get_reports_from_population(self):
...
+        self.assertEqual(sel, Selection([]))
+
+    def test_NodeSetCompound(self):
+        j = '''{"NodeSet0": { "node_id": [1] },
+                "NodeSet1": { "node_id": [2] },
+                "NodeSetCompound": ["NodeSet0", "NodeSet1", "NodeSet2"],
+                "NodeSet2": { "node_id": [3] }
+        }'''
+        sel = NodeSets(j).materialize("NodeSetCompound", self.population)
+        self.assertEqual(sel, Selection(((1, 4), )))
```
matz-e, line 423:
```suggestion
    def test_NodeSetCompound(self):
        j = '''{"NodeSet0": { "node_id": [1] },
                "NodeSet1": { "node_id": [2] },
                "NodeSetCompound0": ["NodeSet0", "NodeSet1"],
                "NodeSetCompound1": ["NodeSetCompound0", "NodeSet2"],
                "NodeSet2": { "node_id": [3] }
        }'''
        sel = NodeSets(j).materialize("NodeSetCompound1", self.population)
        self.assertEqual(sel, Selection(((1, 4), )))
```

src/node_sets.cpp
```diff
@@ -0,0 +1,320 @@
...
+                    const std::map<std::string, CompoundTargets>& compound_rules,
+                    size_t depth) {
+    if (depth > MAX_COMPOUND_RECURSION) {
+        throw SonataError("Compound node_set recursion depth exceeded");
+    }
+
+    for (auto const& it : compound_rules) {
+        for (auto const& name : it.second) {
+            if (compound_rules.count(name) > 0) {
+                check_compound(node_sets, compound_rules, depth + 1);
```
matz-e, line 229: I feel like I'm missing something here. How I read this, the recursion is called with the same parameters as before, but depth increased by one. To me, this would trigger as soon as any compound set refers to another one. See my test suggestion above.

Shouldn't we linearly iterate over the compound sets, and try to recursively resolve each one until max depth reached or loop detected?

src/node_sets.cpp
matz-e: I vote for `summon`.

src/node_sets.cpp
```diff
@@ -0,0 +1,320 @@
...
+        ret += fmt::format(R"(  "{}": {{)", pair.first);
+        for (const auto& pred : pair.second) {
+            ret += pred->toJSON();
+            ret += ", ";
+        }
+        replace_trailing_coma(ret, ' ');
+        ret += "},\n";
+    }
+    replace_trailing_coma(ret, '\n');
+    ret += "}";
```
matz-e, line 314: JSON parsing is already included by an external library, why are we not using that instead of manually assembling all this?

src/node_sets.cpp
mgeplf: Nope, you're 100% right; great eyes.  Fixed and added test case.

> until max depth reached or loop detected?

I'm doing the simple max depth reached for now; I expect this case to be rare, so only want to throw a warning without being too specific.

src/node_sets.cpp
mgeplf: I was using it originally to serialize, but it got ugly (IMO).  The `toJSON` is mainly to help w/ debugging/testing, so it's valuable there, but I don't know how much usage it'll get 'in the real world'.

matz-e: Fair, the code is pretty simple, too. Maybe a comment on that?

src/node_sets.cpp
matz-e: Thank you, new code looks good to me. Limits can be bumped if needed, and loop detection indeed seems like overkill here.

include/bbp/sonata/node_sets.h
```diff
@@ -57,7 +53,7 @@ class SONATA_API NodeSets
     std::string toJSON() const;
 
   private:
-    std::map<std::string, NodeSetRules> node_sets_;
+    std::unique_ptr<detail::NodeSets> impl_;
```
matz-e, line 56: Are we still doing the pimple interface thing?

matz-e (dismissed):
LGTM, would be nice to have this in a release

mgeplf:
@matz-e fixed up the review, everything is passing now.  Are you reviewing on behalf of HPC?

tomdele (approved):

tomdele:
Looks good ! 
Using the node sets in the simulation will be such a relief.

matz-e (approved):

PR #108, opened 2020-11-02
Fix for a performance issue with writers initialization

sergiorg-hpc:


src/reports/library/implementation_interface.hpp
```diff
@@ -103,7 +103,7 @@ struct ParallelImplementation {
             mpi_size_type = MPI_UINT32_T;
         }
 
-        int global_rank, global_size;
+        int global_rank, global_size, local_size;
```
sergiorg-hpc, line 106: Not strictly needed, but better to change it to `local_size` instead.

jorblancoa (approved):

pramodk:
Are we tagging new version and deploying this on BB5?

jorblancoa:
We should do that yes.
@sergiorg-hpc could you create a spack PR with a new version in libsonata-report?
Just replace existing one for 0.1.3

sergiorg-hpc:
> We should do that yes.
> @sergiorg-hpc could you create a spack PR with a new version in libsonata-report?
> Just replace existing one for 0.1.3

Perfect, here's the PR for that:
https://github.com/BlueBrain/spack/pull/945

PR #109, opened 2020-11-11
Wheel ci

tomdele:


wizmer (dismissed):

wizmer:
On attend le resultat de Travis avant de merge ?

tomdele:
yeah

.github/workflows/upload-wheels.yaml
```diff
@@ -0,0 +1,96 @@
+name: Build
+
+on:
+  push:
+    tags:
+      - 'v[0-9]+.[0-9]+.[0-9]+'
+
+
+env:
+  CIBW_MANYLINUX_X86_64_IMAGE: manylinux2014
```
mgeplf, line 10: why manylinux2014?

.github/workflows/upload-wheels.yaml
```diff
@@ -0,0 +1,96 @@
...
+          CIBW_BEFORE_BUILD: 'brew update; brew install cmake hdf5'
+        run: |
+          python -m cibuildwheel --output-dir dist
+      - name: Store wheel as artifact
+        uses: actions/upload-artifact@v2
+        with:
+          name: dist
+          path: dist
+
+  build_sdist:
```
mgeplf, line 58: if we keep the `python_push_sdist.sh` file, it makes things easier to test, imo, and move away to another CI system at somepoint.  Does adding it in the action make something easier?

wizmer:
Just to let you know @mgeplf this PR is out of date. This has already been merged using the template.
But wrt/ your comments.
- yes, I'll try to use manylinux 2010 asap. This was probably a mistake.
- The new way have several advantages:
  * it build the sdist and uploads it as an artifact which is then available from the Github UI.
  * the final upload to pypi will be triggered only if both tasks `build_sdist` and `build_wheels` have succeeded uploading the artifacts. That means we won't end-up with the sdist being pushed while the wheels are failing.
  * The final task that upload to pypi uses the github secret that contains the pypi token, this seems to be the standard way to upload to pypi from github actions. Since it is using a token instead of the pypi password, it is probably safer.

wizmer:
This can be closed, no ?

mgeplf:
Yeah, let's close it.

PR #110, opened 2020-11-20
Build also python27 wheel for libsonata

wizmer:


pramodk:
@wizmer : Is providing Python 2.7 wheel considered as promoting Python 2? :grin:

wizmer:
LOL. Unfortunately, as long as bluepy supports python 2.7, I have no other choice than to build these wheels...
But don't worry, it seems that the fact that nexus-forge is python 3 only was the providential argument we needed to change the mentalities and bluepy should drop python 2 support in the following weeks.
:crossed_fingers:

pramodk:
>  But don't worry, it seems that the fact that nexus-forge is python 3 only was the providential argument we needed to change the mentalities and bluepy should drop python 2 support in the following weeks.

Good to hear! We can remember this covid-19 time for getting rid of Python2 from bluepy! :grin:

matz-e (commented):
But why if we're just going to drop it in a few weeks?

wizmer:
Bluepy currently needs a bug fix because of h5py>=3 that I can't wait a few weeks before merging.

.github/workflows/publish-sdist-wheels.yml
```diff
@@ -10,7 +10,7 @@ env:
   MACOSX_DEPLOYMENT_TARGET: '10.13'
   CIBW_BUILD_VERBOSITY: 2
   # Only build on Python 3.x
-  CIBW_BUILD: 'cp3?-*'
+  CIBW_BUILD: 'cp3?-* cp27-*'
```
tomdele, line 13: I tried before and this fails due to cibuildwheel version compatibility problems when running with 1.6.4. 
You should relax the version number to make it works probably. But this is a bit scary to have wheels made with different versions ....

wizmer: Ok.....
Let's drop python 2 instead...

wizmer:
Bad idea

PR #112, opened 2020-12-16
Fix python bindings for Selection with negative values

### LINKED ISSUE #94: Selection behavior unsafe with negative numbers
The Selection object can lead to non consistent behaviors if you provide a negative number/range: 
```
Selection([-1, 0, 1, 2]).flatten()
array([], dtype=uint64)

Selection([0, 1, 2, -1]).flatten()
array([0, 1, 2], dtype=uint64)

Selection([0, 1, -1, 2]).flatten()                                                                                                                                                                 array([0, 1, 2], dtype=uint64)

Selection([[-2, -1], [0, 1], [4,6]]).flatten()
ValueError: array has incorrect number of dimensions: 2; expected 1

Selection([[-1, 2], [4, 4], [5,6]]).flatten()
ValueError: array has incorrect number of dimensions: 2; expected 1
```
Same for `Selection.ranges`.

mgeplf:
* use new environment setting post CVE-2020--15228
* fixes #94


tomdele (approved):

tomdele:
Unfortunately the C++ part is not fixed but I think this is still good to have this.

python/bindings.cpp
```diff
@@ -375,13 +375,45 @@ PYBIND11_MODULE(_libsonata, m) {
...
+                     if (raw[i] < 0) {
+                         throw SonataError("Negative value passed to Selection");
+                     }
+                 }
+
+                 return Selection::fromValues(raw.data(0), raw.data(raw.shape(0)));
+             }),
              "values"_a,
-             "Selection from list of IDs")
+             "Selection from list of IDs: passing np.array with dtype np.uint64 is faster")
```
matz-e, line 416: Is this quantifiable?

mgeplf:
> Unfortunately the C++ part is not fixed but I think this is still good to have this.

How so?  It should be strongly typed to only accept `uint64_t`.

python/bindings.cpp
mgeplf: Wouldn't have written it otherwise:
```
In [1]: import numpy as np
In [2]: import libsonata
In [3]: uint64 = np.arange(1000000, dtype=np.uint64)
In [4]: int64 = np.arange(1000000, dtype=np.int64)
In [5]: a = range(1000000)
In [6]: %timeit libsonata.Selection(uint64)
2.07 ms ± 11.2 µs per loop (mean ± std. dev. of 7 runs, 100 loops each)
In [7]: %timeit libsonata.Selection(int64)
20.3 ms ± 68.7 µs per loop (mean ± std. dev. of 7 runs, 10 loops each)
In [8]: %timeit libsonata.Selection(a)
143 ms ± 525 µs per loop (mean ± std. dev. of 7 runs, 10 loops each)
```

pybind11's conversion machinenery kicks in when the types don't match, causing a conversion and thus the overhead.

matz-e: nice!

mgeplf:
> > Unfortunately the C++ part is not fixed but I think this is still good to have this.
> 
> How so? It should be strongly typed to only accept `uint64_t`.

@tomdele ping

tomdele:
> > > Unfortunately the C++ part is not fixed but I think this is still good to have this.
> > 
> > 
> > How so? It should be strongly typed to only accept `uint64_t`.
> 
> @tomdele ping

Whao I think I was quite tired ... Indeed everything is based on `Value` which is a `uint64_t`.  I guess I just checked the C++ tests and I have not seen any "negative --> throw" check. 
So this all good on my side.

PR #116, opened 2021-01-07
remove travis, switch to github actions for tests

mgeplf:
* tests first run the CPP w/ clang & gcc, then the python ones, then the
  macos ones
* only support running tests on py3


.github/workflows/run-tests.yml
```diff
@@ -0,0 +1,81 @@
...
+        run: sudo apt-get update && sudo apt-get install libhdf5-dev doxygen
+
+      - name: Build and run unittests
+        run: |
+            ./ci/python_test.sh
+
+  build-macos:
+    name: Run tests on macos
+    runs-on: macos-latest
+    needs: build-linux-python # don't bother with mac until linux passes
```
wizmer, line 62: Imo, the shorter the CI duration the better for the devs. I don't see the point of this.

mgeplf: I assume they'll throttle number of runners at some point.
Do you know if they send an email as soon as one job fails, or only after all have finished?

wizmer: If A and B are running in parallel, if A fails, it will abort B so I guess the mail is sent asap.

Source:
 The Linux build https://github.com/BlueBrain/MorphIO/runs/1656377247?check_suite_focus=true#step:5:2575
which was aborted because the Mac OS build failed (because of the exact same issue you solved in this PR)

wizmer (approved):
Up to you for the `needs: build-linux-python`

Other than that: LGTM

.github/workflows/run-tests.yml
mgeplf: cool, good to know, thanks.

wizmer (approved):

.github/workflows/run-tests.yml
```diff
@@ -0,0 +1,79 @@
...
+    steps:
+      - name: Checkout repository
+        uses: actions/checkout@v2
+        with:
+          submodules: 'true'
+
+      - name: Install packages on MacOS
+        run: |
+            brew update
+            brew unlink gcc@8 gcc@9  # work around for https://github.com/actions/virtual-environments/issues/2391
```
wizmer, line 71: You also need to fix https://github.com/BlueBrain/libsonata/blob/master/.github/workflows/publish-sdist-wheels.yml#L48

mgeplf: Indeed, I will do that in a separate review.

PR #118, opened 2021-01-15
lookup vector of values in matchAttributeValues

mgeplf:
* when several values are to be tested against the whole vector,
  only load the vector once, and compare each instance against the
  wanted list

* for nodeset with multiple values:
    ex:
        "L23": {
           "mtype": [ "L23_BP", "L23_BTC", "L23_CHC", "L23_DBC", "L23_LBC",
                      "L23_MC", "L23_NBC", "L23_NGC", "L23_SBC", "L2_IPC",
                      "L2_TPC:A", "L2_TPC:B", "L3_TPC:A", "L3_TPC:C"]
        }
    The old code takes about ~78ms to match 1177860 of 4234929 elements,
    and the new takes ~63ms.

* for nodeset with a single value:
    ex:
        "L23_BP": { "mtype": "L23_BP" }
    The old code takes about ~7ms to match 9599 of 4234929 elements,
    and the new takes ~8ms.

    and:
        "EXC": { "synapse_class": "EXC" }
    The old code takes about ~15ms to match 3738444 of 4234929 elements,
    and the new takes ~16ms.

### DIFF
--- a/src/nodes.cpp
+++ b/src/nodes.cpp
@@ -27,16 +27,26 @@ NodePopulation::NodePopulation(const std::string& h5FilePath,
 
 
 namespace {
+
 template <typename T>
-Selection _getMatchingSelection(const std::vector<T>& values, const T value) {
+Selection _getMatchingSelection(const std::vector<T>& values, const std::vector<T>& wanted) {
     Selection::Values idx;
     Selection::Value id = 0;
 
-    for (const auto& v : values) {
-        if (v == value) {
-            idx.push_back(id);
+    if (wanted.size() == 1) {
+        for (const auto& v : values) {
+            if (v == wanted[0]) {
+                idx.push_back(id);
+            }
+            ++id;
         }
-        ++id;
+    } else {
+        for (const auto& v : values) {
+            if (std::find(wanted.cbegin(), wanted.cend(), v) != wanted.cend()) {
+                idx.push_back(id);
+            }
+            ++id;
+        }
     }
     return Selection::fromValues(idx);
 }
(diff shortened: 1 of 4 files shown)


src/nodes.cpp
```diff
@@ -65,31 +76,55 @@ Selection NodePopulation::matchAttributeValues(const std::string& name, const T
...
+Selection NodePopulation::matchAttributeValues<std::string>(
+    const std::string& name, const std::vector<std::string>& values) const {
     if (enumerationNames().count(name) > 0) {
         const auto enum_values = enumerationValues(name);
-        const auto pos = std::find(enum_values.begin(), enum_values.end(), value);
-        if (pos == enum_values.end()) {
+        std::vector<size_t> wanted_enum_value;
+        wanted_enum_value.reserve(values.size());
+        for (const auto& v : values) {
+            const auto wanted_index = std::find(enum_values.cbegin(), enum_values.cend(), v);
```
matz-e, line 93: So worst case here is `n log(n)` string comparisons if we want to match close to all values… would a look-up table be warranted? That would give `2n` with a bunch of additional memory consumption.

src/nodes.cpp
```diff
@@ -27,36 +27,47 @@ NodePopulation::NodePopulation(const std::string& h5FilePath,
...
+    if (wanted.size() == 1) {
+        for (const auto& v : values) {
+            if (v == wanted[0]) {
+                idx.push_back(id);
+            }
+            ++id;
+        }
+    } else {
+        for (const auto& v : values) {
+            if (std::find(wanted.cbegin(), wanted.cend(), v) != wanted.cend()) {
```
matz-e, line 45: See comment below, maybe a lookup table would be required at some point?

src/nodes.cpp
```diff
@@ -65,31 +76,55 @@ Selection NodePopulation::matchAttributeValues(const std::string& name, const T
...
+    return matchAttributeValues<T>(name, values);
+}
+
 template <>
-Selection NodePopulation::matchAttributeValues<std::string>(const std::string& name,
-                                                            const std::string value) const {
-    // enum attribute, need to look up in the @library
+Selection NodePopulation::matchAttributeValues<std::string>(
+    const std::string& name, const std::vector<std::string>& values) const {
     if (enumerationNames().count(name) > 0) {
```
matz-e, line 88: For enumeration values, this would be a *really convenient* location to do some glob resolving :)

matz-e (commented):
Could you please remind me why we don't allow for `glob` style matching, akin to what we have in the XML recipe? I think that would be a really valuable addition!

mgeplf:
Are all the comments glob related?
Sadly, globs aren't part of the nodeset description.  *however*, I would think they would be a good addition, just not for the this PR.
We can use them internally in BBP, and make a PR to the spec, to add them.

However, that should be debated somewhere else; feel free to open an issue (my main question would be glob vs. regex [insert xkcd here]).

matz-e:
> Are all the comments glob related?
> Sadly, globs aren't part of the nodeset description. _however_, I would think they would be a good addition, just not for the this PR.
> We can use them internally in BBP, and make a PR to the spec, to add them.
> 
> However, that should be debated somewhere else; feel free to open an issue (my main question would be glob vs. regex [insert xkcd here]).

I'll open a NSETM issue then? The look-up table comments are generic! There could be a threshold to switch from `find` in an array to a `O(1)` lookup in a hash map…

mgeplf:
> I'll open a NSETM issue then?

a libsonata one, methinks.

> There could be a threshold to switch from find in an array to a O(1) lookup in a hash map…
My gut is that for the low cardinality we'll be seeing, it probably won't matter, but I will check.

matz-e (approved):
Comments sufficiently answered :)

PR #120, opened 2021-01-18
use binary search on `_getMatchingSelection`

mgeplf:
* speeds up `L23` query from #118 from 63ms -> 50ms


mgeplf:
copy + sort + binary_search seems worth it

matz-e (approved):
Looks good :)

PR #121, opened 2021-01-20
Fix sdist

### LINKED ISSUE #117: libsonata tarball should be compilable
I think the libsonata tarball is bundled without its submodule which makes it uncompilable:

```bash
    -- Found HDF5: /usr/lib64/libhdf5.so;/usr/lib64/libsz.so;/usr/lib64/libz.so;/usr/lib64/libdl.so;/usr/lib64/libm.so (found version "1.8.12")
    CMake Error at extlib/CMakeLists.txt:9 (add_subdirectory):
      add_subdirectory given source "fmt" which is not an existing directory.
    
    
    CMake Error at python/CMakeLists.txt:4 (add_subdirectory):
      add_subdirectory given source "pybind11" which is not an existing
      directory.
    
    
    CMake Error at python/CMakeLists.txt:9 (pybind11_add_module):
      Unknown CMake command "pybind11_add_module".
```

### reproduction

```pip install --no-binary libsonata libsonata```

### PR DESCRIPTION
**Pull request by mgeplf**
* fixes regression from https://github.com/BlueBrain/libsonata/commit/9aba1ea5cde1479771b9e7118d149ef4f39ac2f9
* fixes #117

### CODE REVIEW

File: ci/python_build_sdist.sh [hunk 1]
```diff
@@ -0,0 +1,27 @@
+#!/usr/bin/env bash
+
+# build the sdist file, and try installing it
+
+set -euo pipefail
+
+BASE=$(pwd)/build/
+
+mkdir -p $BASE
```
matz-e, line 9: Why is it `"$VENV"`, but `$BASE`? `"$(pwd)"` could include spaces…

matz-e (dismissed):

ci/python_build_sdist.sh
mgeplf: 100% right, thanks.

matz-e (approved):
Looks better :)

PR #129, opened 2021-02-03
Trigger PRs on `pull_request` trigger

wizmer:


mgeplf:
what happens if you enble actions on your fork?

wizmer:
They are already allowed.

mgeplf:
They were allowed when you made a push?

wizmer:
Yes. You mean in settings -> Actions, right ? Yes, they were.
I don't understand why the checks are still listed up there even though they are not being triggered.

mgeplf:
> I don't understand why the checks are still listed up there even though they are not being triggered.

My only guess is b/c they are listed as `Required`?

wizmer:
Ok, and now that I have changed the trigger `docstring_check` is being triggered.

.github/workflows/docstring_check.yaml
```diff
@@ -1,6 +1,8 @@
 name: docstring-check
+
 on:
-    push:
+    pull_request:
```
mgeplf, line 4: I guess it could also be `[pull_request, push]` for the best of both worlds?

wizmer: I'm not sure keeping push adds value but it's debatable I guess. Also the jobs that list both pull_request and push will get triggered twice if you make a PR from a branch on BlueBrain remote.

mgeplf:
> I'm not sure keeping push adds value but it's debatable I guess. 

I use it for WIP branches, before I make the PR, such that the PR doesn't fail testing.

wizmer:
I see, I see, thanks for explaining.

mgeplf:
any idea why coverage isn't working?

Also, will there be any interaction w/ the release process, do you think?

wizmer:
Because I kept the coverage just on push since it is not being tested but it is rather informational.
And no that should not change anything.

tomdele:
And we don t have the problem we had with the double run of the CI at every push or merger (I cannot remember...) ?

wizmer:
We will have it now I think but only for PRs made on a BlueBrain branch

wizmer:
I think I need to merge this because it seems there is a bug with coverage-test still expected even when I remove completely the file.

mgeplf:
> I think I need to merge this because it seems there is a bug with coverage-test still expected even when I remove completely the file.

Could you wait until you have other's reviews before merging?

> Only trigger coverage on master push

Why only on master? I explained above how it's useful to see it on other branches as well.

wizmer:
> Could you wait until you have other's reviews before merging?

I was trying to understand why coverage is not working as intended and needed the merge for this.

wizmer:
> Why only on master? I explained above how it's useful to see it on other branches as well.

Where ? You explained that it is useful having the CI running because you can see where the tests are failing. This is not related to coverage. But feel free to change the behavior anyway, I was just trying to help.

mgeplf:
> I was trying to understand why coverage is not working as intended and needed the merge for this.

I still think it's helpful to let others have a chance to review and comment.

> This is not related to coverage. 

I consider coverage as part of CI (like the docstring tests, and formatting); if it's going down, I want to know what I'm missing - sorry I should have been more clear.

wizmer:
FYI: The coverage is still showing as expected because of the repository settings.

Daniel explained that to me this morning for an unrelated question: https://bbpteam.epfl.ch/project/issues/browse/HELP-13549

mgeplf:
I don't follow; the thing that changed w/ coverage is that it's only on the master branch, so I would expect it to still show up and run on master pushes, and not run on non-master branches.

Are you referring to https://github.com/BlueBrain/libsonata/pull/129#issuecomment-772346837 ?

wizmer:
As of today, if a PR is made, it won't trigger the coverage checks but those checks are still required for the PR to be "green". This is an inconsstency and I was wondering why. As for NeuroR, I was wondering why Travis was still listed in the list of required checks. The issue is in fact not related to this PR. (in fact before this PR the issue was even more present since the CI checks were required but there was no trigger). 

So one need sync the triggers of the PR and the required items in the repo options. Either coverage is required on PRs and we need to fix the github action, or it is not required and we need to fix the repo settings.

mgeplf:
> As of today, if a PR is made, it won't trigger the coverage checks but those checks are still required for the PR to be "green".

Yeah, and I can't merge without overriding the required checks, which is a regression.

Coverage is required; it also seems that now we're also running each job twice.

PR #131, opened 2021-02-03
Fix include from nlohmann json

alkino:
The include file is nlohmann/json.hpp, let it include locally so it can works
when building from external dependencies


extlib/CMakeLists.txt
```diff
@@ -11,7 +11,7 @@ add_subdirectory(fmt)
 add_library(nlohmann_json INTERFACE)
 add_library(nlohmann_json::nlohmann_json ALIAS nlohmann_json)
 target_include_directories(nlohmann_json INTERFACE
-    nlohmann
+    .
```
matz-e, line 14: I think this is a bit broader than desired? I'd nest the `nlohmann` directory more.

alkino: I agree ^^

mgeplf:
Cool, thanks for doing this; was hoping to do exactly the same thing, so you've saved me time.

I'll try and figure out why format isn't working properly.

alkino:
Your fix is really good ;)

matz-e (approved):

alkino:
@mgeplf I cannot merge it, only authorized users can. Are you of one them?

mgeplf:
Currently CI is broken, because of this: https://github.com/BlueBrain/libsonata/pull/129 and I haven't had a chance to fix it.
Since bb5 is down tomorrow, it's on my list for then.

PR #134, opened 2021-02-10
have CI work for internal/external PRs and pushes

mgeplf:
* 'external' PRs (ie: not part of the BlueBrain org) run CI
* avoid running CI twice for 'internal' PRs
* have CI run on all pushes

### DIFF
--- a/.github/workflows/clang_format_check.yaml
+++ b/.github/workflows/clang_format_check.yaml
@@ -5,6 +5,9 @@ jobs:
   build:
     name: clang-format-check
     runs-on: ubuntu-18.04
+    # Run on external PRs, but not internal PRs as they'll be run by the push
+    if: github.event_name == 'push' || github.event.pull_request.head.repo.full_name != github.repository
+
     steps:
         - name: Fetch repository
           uses: actions/checkout@v1
--- a/.github/workflows/coverage_test.yaml
+++ b/.github/workflows/coverage_test.yaml
@@ -1,13 +1,13 @@
 name: coverage-test
-on:
-    push:
-      branches:
-        - master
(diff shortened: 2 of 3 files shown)


wizmer (approved):
You can also put the coverage as `required` in the repo settings if you want but I think it won't change anything because the coverage cannot return False in its current implem (if I am not mistaken, there is not coverage threshold)

tomdele:
I am still puzzled we have to do something so complex. I spent some time to check but couldn't find anything straightforward (that s why my first version was for push only).

tomdele (approved):

wizmer:
If you use the forking workflow you can simply use: `pull_request`.
If everyone works on the BlueBrain remote, you can simply use `push`.
Otherwise, you do this I guess.

mgeplf:
Huh, I had taken you guys off b/c I thought the `Skipped` ones were marked as `Required`, but apparently that's ok...

> You can also put the coverage as required in the repo settings if you want but I think it won't change anything because the coverage cannot return False in its current implem (if I am not mistaken, there is not coverage threshold)

Yeah, good point. I thought it was - it's true that it can't return false, but at least we don't want it to fail.

> I am still puzzled we have to do something so complex. I spent some time to check but couldn't find anything straightforward (that s why my first version was for push only).

Yeah, it seems weird that there isn't a default for this, b/c it seems like a common use case.  I think I had the original clang-checks as being push only, so prob. may fault - plus we've only had `internal` contributions thus far.

> [...] Otherwise, you do this I guess. 

Sad but true.

mgeplf:
still getting duplicate runs, will have to investigate further

PR #136, opened 2021-02-11
FMT and HighFive are targets, depends directly of it

alkino:


extlib/CMakeLists.txt
```diff
@@ -1,18 +1,12 @@
...
-add_library(HighFive INTERFACE)
-target_include_directories(HighFive INTERFACE
-    HighFive/include/
-    ${HDF5_INCLUDE_DIRS})
-target_link_libraries(HighFive INTERFACE
-    ${HDF5_C_LIBRARIES})
-
+option(HIGHFIVE_USE_BOOST "Disable boost in HighFive" OFF)
+add_subdirectory(HighFive)
+option(FMT_INSTALL "Install fmt for static library" ON)
```
matz-e, line 3: do we really want this? are these options exposed on, e.g., to the user via `ccmake`?

alkino: Yes, users can modify it.

alkino (commented):
Bump

alkino:
@jorblancoa ?

extlib/CMakeLists.txt
matz-e: I'm not convinced that we should install stuff from submodules. That really smells like dependency conflicts down the line.

alkino: The question here is: Should we install fmt even if it's a header-only depends?

alkino: The answer is yes. Otherwise there is no target available for cmake.

matz-e: Errm, why do we need the target? Shouldn't this be a private thing for libSONATA?

alkino:
Target is the new way to make depends between things in cmake.

It add includes, libraries and builds variable transparently. More over it is the simple way to make transactional depends.

matz-e:
> Target is the new way to make depends between things in cmake.
> 
> It add includes, libraries and builds variable transparently. More over it is the simple way to make transactional depends.

Shouldn't private targets not be needed after installation, though? Thus FMT should not have to be installed by itself?

alkino:
The name is misleading, it will not be installed

CMakeLists.txt
```diff
@@ -115,6 +115,10 @@ foreach(TARGET sonata_shared sonata_static)
     target_compile_options(${TARGET}
         PRIVATE ${SONATA_COMPILE_OPTIONS}
     )
+    target_link_libraries(${TARGET}
+        PRIVATE fmt::fmt-header-only
+        PUBLIC HighFive
```
matz-e, line 120: Why did this go from PRIVATE to PUBLIC?

alkino: Because of: https://github.com/BlueBrain/libsonata/blob/master/include/bbp/sonata/report_reader.h#L9

matz-e: Could that not be forward-declared? :(

alkino: not easily, can we do this in a new PR?
I can create a ticket for that

matz-e:
> The name is misleading, it will not be installed

Indeed, I did not see any additionally installed files. So from my side this would be "OK", even if I'm cringing.

@mgeplf I'll let you have a word, I think I'm done scrutinizing.

extlib/CMakeLists.txt
mgeplf: I'm not following this; why does someone building the project need to have options for our internal dependencies?

CMakeLists.txt
mgeplf: I think we should do that first, so that HighFive doesn't bleed through the public headers; then we don't need to make it `PUBLIC`, which should be avoided, imo.

mgeplf (commented):
I don't understand the:

> This way mpicc and mpicxx is no more needed.

In the commit message, could you expand on that?

extlib/CMakeLists.txt
alkino: The option here is use to set the option, not to expose them (they already are exposed)

CMakeLists.txt
alkino: ok :(

alkino:
I don't really remember about mpicc mpicxx, reading the patch I guess it's useless now.

CMakeLists.txt
alkino: This is not an easy patch, and today highfive is already a public dependency, but hidden (people have to guess it ^^)

mgeplf: Good point.

However, isn't it a matter of wrapping `ReportReader::file_` in `std::unique_ptr`, and saving the population name instead of `pop_group_`?

alkino: Good idea, i was trying to put `pop_group_` inside `std::unique_ptr` too, it was leading to complicate things ^^

alkino: It implies to store the file in the population class... That is the same, do we really want a shared_ptr?

PR #137, opened 2021-02-24
Remove code duplicate inside libsonatareport

alkino:
Now, that a split has been asked and done for reports inside BlueBrain/libsonatareport, we can remove this part from libsonata.

You have to depends of libsonatareport for still having this code.


mgeplf:
Is there a 'go-with' PR that keeps the interface the same within libsonata?

When @pramodk and I discussed this, the idea was that libsonata would continue to have the same API, and the build wouldn't become more complicated.  It seems like all the report reading capabilites are gone, if I"m reading this right, which means that, for instance, bluepysnap & the viz team are losing access to reports - which is something that I really want to avoid.

pramodk:
> You have to depends of libsonatareport for still having this code.

@alkino : could you describe a bit : 

* when you say depends on libsonatareport, is that in the code as submodule or at runtime?
* currently users are using libsonata (on BB5 or via pip installation), how they would be use this considering this split?

pramodk:
@mgeplf : Ooops...I wrote the questions at the same time. Based on our discussion in the past, we will try to get clarify based on use cases and users expectations. 

I have converted this PR to draft until we have this clarity.

mgeplf:
> I wrote the questions at the same time.

No problem.

> I have converted this PR to draft until we have this clarity.

Ok; let me know if you have any questions.

alkino:
Endly we will not do that.

reader in libsonata, writer in libsonatareport

PR #142, opened 2021-04-12
WIP circuit_config.json implementation

mgeplf:
* Based off Brion version

### DIFF
--- a/.gitmodules
+++ b/.gitmodules
@@ -10,3 +10,6 @@
 [submodule "extlib/Catch2"]
 	path = extlib/Catch2
 	url = https://github.com/catchorg/Catch2.git
+[submodule "extlib/filesystem"]
+	path = extlib/filesystem
+	url = https://github.com/gulrak/filesystem/
--- a/include/bbp/sonata/config.h
+++ b/include/bbp/sonata/config.h
@@ -26,11 +26,28 @@
 namespace bbp {
 namespace sonata {
 
-struct SONATA_API PopulationProperties
-{
+/**
+ * Stores pouplation-specific network information.
+ */
+struct SONATA_API PopulationProperties {
+    /**
+     * Population type
+     */
     std::string type;
+
+    /**
+     * Path to the template HOC files defining the E-Mode
+     */
     std::string biophysicalNeuronModelsDir;
+
+    /**
+     * Path to the directory containing the morphologies
+     */
     std::string morphologiesDir;
+
+    /**
+     * Dictionary for alternate directory paths.
+     */
     std::map<std::string, std::string> alternateMorphologyFormats;
 };
 
@@ -40,15 +57,10 @@ struct SONATA_API PopulationProperties
 class SONATA_API CircuitConfig
 {
   public:
-
     /**
-     * @brief Parses a SONATA JSON Config file and checks it complies with the specification.
+     * Parses a SONATA JSON config file.
      *
-     * @param contents string containing the JSON content of the file
-     * @param basePath string the path to the directory where the original file is located in
-     *                 on disk. It is needed for variable expansion.
-     *
-     * @throws SonataError on:
+     * \throws SonataError on:
      *          - Ill-formed JSON
      *          - Missing mandatory entries (in any depth)
      *          - Missing entries which become mandatory when another entry is present
@@ -61,10 +73,10 @@ class SONATA_API CircuitConfig
     ~CircuitConfig();
 
     /**
-     * @brief Loads a SONATA JSON config file from disk and returns a CircuitConfig object which
-     *        parses it.
+     * Loads a SONATA JSON config file from disk and returns a CircuitConfig object which
+     * parses it.
      *
-     * @throws SonataError on:
+     * \throws SonataError on:
      *          - Non accesible file (does not exists / does not have read access)
      *          - Ill-formed JSON
      *          - Missing mandatory entries (in any depth)
@@ -74,81 +86,57 @@ class SONATA_API CircuitConfig
     static CircuitConfig fromFile(const std::string& path);
 
     /**
-     * @brief Returns the path to the node sets file.
-     *
-     * @returns a string with the path to the node sets file, or an empty string if none
+     * Returns the path to the node sets file.
      */
     std::string getNodeSetsPath() const;
 
     /**
-     * @brief Returns a set with all available population names across all the node networks.
-     *
-     * @returns a std::set with population names.
+     *  Returns a set with all available population names across all the node networks.
      */
     std::set<std::string> listNodePopulations() const;
 
     /**
-     * @brief Creates and returns a NodePopulation object, initialized from the given population,
-     *        and the node network it belongs to.
+     * Creates and returns a NodePopulation object, initialized from the given population,
+     * and the node network it belongs to.
      *
-     * @param name Name of a node population.
-     *
-     * @returns A NodePopulation object.
-     *
-     * @throws SonataError if the given population does not exist in any node network.
+     * \throws SonataError if the given population does not exist in any node network.
      */
     NodePopulation getNodePopulation(const std::string& name) const;
 
     /**
-     * @brief Returns a set with all available population names across all the edge networks.
-     *
-     * @returns a std::set with population names.
+     * Returns a set with all available population names across all the edge networks.
      */
     std::set<std::string> listEdgePopulations() const;
 
     /**
-     * @brief Creates and returns an EdgePopulation object, initialized from the given population,
-     *        and the edge network it belongs to.
+     * Creates and returns an EdgePopulation object, initialized from the given population,
+     * and the edge network it belongs to.
      *
-     * @param name Name of an edge population.
-     *
-     * @returns An EdgePopulation object.
-     *
-     * @throws SonataError if the given population does not exist in any edge network.
+     * \throws SonataError if the given population does not exist in any edge network.
      */
     EdgePopulation getEdgePopulation(const std::string& name) const;
 
     /**
-     * @brief Return a structure containing node population specific properties, falling
-     *        back to network properties if there are no population-specific ones.
+     * Return a structure containing node population specific properties, falling
+     * back to network properties if there are no population-specific ones.
      *
-     * @param name Nam of a node population.
-     *
-     * @return The properties of the given population.
-     *
-     * @throws SonataError if the given population name does not correspond to any existing
+     * \throws SonataError if the given population name does not correspond to any existing
      *         node population.
      */
     PopulationProperties getNodePopulationProperties(const std::string& name) const;
 
     /**
-     * @brief Return a structure containing edge population specific properties, falling
-     *        back to network properties if there are no population-specific ones.
+     * Return a structure containing edge population specific properties, falling
+     * back to network properties if there are no population-specific ones.
      *
-     * @param name Nam of an edge population.
-     *
-     * @return The properties of the given population.
-     *
-     * @throws SonataError if the given population name does not correspond to any existing
+     * \throws SonataError if the given population name does not correspond to any existing
      *         edge population.
      */
     PopulationProperties getEdgePopulationProperties(const std::string& name) const;
 
     /**
-     * @brief Returns the configuration file JSON whose variables have been expanded by the
-     *        manifest entries.
-     *
-     * @returns a JSON string with the expanded configuration file.
+     * Returns the configuration file JSON whose variables have been expanded by the
+     * manifest entries.
      */
     std::string getExpandedJSON() const;
 
--- a/include/bbp/sonata/config.h
+++ b/include/bbp/sonata/config.h
@@ -1,0 +1,90 @@
+/*************************************************************************
+ * Copyright (C) 2018-2020 Blue Brain Project
+ *                         Jonas Karlsson <jonas.karlsson@epfl.ch>
+ *                         Juan Hernando <juan.hernando@epfl.ch>
+ *
+ * This file is part of 'libsonata', distributed under the terms
+ * of the GNU Lesser General Public License version 3.
+ *
+ * See top-level COPYING.LESSER and COPYING files for details.
+ *************************************************************************/
+
+#pragma once
+
+#include <memory>  // std::unique_ptr
+#include <set>
+#include <string>
+#include <vector>
+
+#include <bbp/sonata/edges.h>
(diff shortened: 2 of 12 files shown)


src/config.cpp
```diff
@@ -0,0 +1,322 @@
...
+
+        for (const auto& vI : variables) {
+            const auto& vIKey = vI.first;
+            const auto& vIValue = vI.second;
+
+            for (auto& vJ : variablesCopy) {
+                auto& vJValue = vJ.second;
+                const auto startPos = vJValue.find(vIKey);
+
+                if (startPos != std::string::npos) {
```
matz-e, line 72: Shouldn't we check that the key is followed by something non-alphanumeric to make sure that the found variable is terminated properly?

Seems that actually `${BASE_DIR}` etc would be better than `$BASE_DIR`.

include/bbp/sonata/config.h
```diff
@@ -0,0 +1,90 @@
+/*************************************************************************
+ * Copyright (C) 2018-2020 Blue Brain Project
+ *                         Jonas Karlsson <jonas.karlsson@epfl.ch>
+ *                         Juan Hernando <juan.hernando@epfl.ch>
```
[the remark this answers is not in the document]
mgeplf: yeah, one can't remove copyrights from the header

.gitmodules
```diff
@@ -10,3 +10,6 @@
 [submodule "extlib/Catch2"]
 	path = extlib/Catch2
 	url = https://github.com/catchorg/Catch2.git
+[submodule "extlib/filesystem"]
+	path = extlib/filesystem
+	url = https://github.com/gulrak/filesystem/
```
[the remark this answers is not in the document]
mgeplf: I'd like to steer clear of boost if at all possible.

src/config.cpp
mgeplf: > Shouldn't we check that the key is followed by something non-alphanumeric to make sure that the found variable is terminated properly?

Yeah, good point.  Nice catch.

> Seems that actually ${BASE_DIR} etc would be better than $BASE_DIR.

It's following the spec: https://github.com/AllenInstitute/sonata/blob/master/docs/SONATA_DEVELOPER_GUIDE.md#tying-it-all-together---the-networkcircuit-config-file

include/bbp/sonata/config.h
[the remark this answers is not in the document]
matz-e: isn't this a new file, though?

src/config.cpp
matz-e: I see this just below your link:
```
"network": "${BASE_DIR}/circuit_config.json"
```

Seems that the variable replacement should be more than just plain search + replace.

matz-e: Seems fair.

include/bbp/sonata/config.h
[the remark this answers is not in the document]
matz-e: Also, it's the year 2021… :)

NadirRoGue:
Apologies for the last commit and its revert counterpart.

NadirRoGue:
Pinging @mgeplf @ferdonline @matz-e 
Any thoughts on this?

include/bbp/sonata/config.h
```diff
@@ -25,61 +26,131 @@
...
+/**
+ *  Read access to a SONATA circuit config file.
  */
 class SONATA_API CircuitConfig
 {
   public:
-    /** Load SONATA circuit config JSON
+
+    /**
+     * @brief Parses a SONATA JSON Config file and checks it complies with the specification.
```
mgeplf, line 45: nit: `complies with the specification` not sure how much we want to guarantee we'll do this, so I'd drop that wording,

include/bbp/sonata/config.h
```diff
@@ -25,61 +26,131 @@
...
      */
     EdgePopulation getEdgePopulation(const std::string& name) const;
 
-    /** Return the names of the components available */
-    std::set<std::string> listComponents() const;
+    /**
+     * @brief Return a structure containing node population specific properties, falling
+     *        back to network properties if there are no population-specific ones.
+     *
+     * @param name Nam of a node population.
```
mgeplf, line 125: Nam -> Name

include/bbp/sonata/config.h
```diff
@@ -25,61 +26,131 @@
...
+     * @throws SonataError if the given population name does not correspond to any existing
+     *         node population.
+     */
+    PopulationProperties getNodePopulationProperties(const std::string& name) const;
+
+    /**
+     * @brief Return a structure containing edge population specific properties, falling
+     *        back to network properties if there are no population-specific ones.
+     *
+     * @param name Nam of an edge population.
```
mgeplf, line 138: Nam -> Name

include/bbp/sonata/config.h
```diff
@@ -25,61 +26,131 @@
...
+    /**
+     * @brief Returns the configuration file JSON whose variables have been expanded by the
+     *        manifest entries.
      *
-     * @param name component name
-     * @throw bbp::sonata::SonataError if component not found
+     * @returns a JSON string with the expanded configuration file.
      */
-    std::string getComponent(const std::string& name) const;
+    std::string getExpandedJSON() const;
```
mgeplf, line 153: nice

src/config.cpp
```diff
@@ -26,8 +26,11 @@
 #include "population.hpp"
 #include "utils.h"
 
+#include <iostream>
```
mgeplf, line 29: debug?

src/config.cpp
```diff
@@ -115,16 +118,53 @@ nlohmann::json expandVariables(const nlohmann::json& json,
     return jsonFlat.unflatten();
 }
 
+std::string getJSONValue(const nlohmann::json& json,
```
mgeplf, line 121: what if a string value isn't desired? perhaps template it?

src/config.cpp
```diff
@@ -216,6 +311,50 @@ std::set<std::string> listPopulations(const std::vector<SubnetworkFiles>& networ
...
+template <typename Population>
+void checkDuplicatePopulationNames(const std::vector<SubnetworkFiles>& networkNodes) {
+    using PopulationStorage = bbp::sonata::PopulationStorage<Population>;
+    std::set<std::string> check;
+
+    for (const auto& network : networkNodes) {
+        const auto population = PopulationStorage(network.elements, network.types);
+        for (auto name : population.populationNames()) {
+            if(check.find(name) != check.end())
+                throw SonataError(fmt::format("Duplicate population name '{}'", name));
```
mgeplf, line 317: not sure it's worthwhile, but for debugging, it might be worth saying which files have duplicates

src/config.cpp
```diff
@@ -216,6 +311,50 @@ std::set<std::string> listPopulations(const std::vector<SubnetworkFiles>& networ
...
+            check.insert(name);
+        }
+    }
+}
+
+void checkBiophysicalNodePopulations(
+        const std::vector<SubnetworkFiles>& networkFiles,
+        const std::map<std::string, PopulationProperties>& nodePopulations) {
+
+    using PopulationStorage = bbp::sonata::PopulationStorage<bbp::sonata::NodePopulation>;
```
mgeplf, line 326: `NodeStorage` exists in `nodes.h`

mgeplf (commented):
some small nits; overall looks reasonable to me.

Does it cover the uses cases on the viz side?

I haven't carefully checked the logic, FYI.

include/bbp/sonata/config.h
NadirRoGue: Agree, better be safe. Removed

include/bbp/sonata/config.h
NadirRoGue: Fixed, thanks.

include/bbp/sonata/config.h
NadirRoGue: Fixed, thanks.

src/config.cpp
NadirRoGue: Yes haha!. Removed.

src/config.cpp
NadirRoGue: I used only within the initialization, but in the future it could be useful as well. 
So, template parameterized, thanks!

src/config.cpp
NadirRoGue: Done, thanks!

src/config.cpp
NadirRoGue: Removed `using` statement and replaced with `NodeStorage`, thanks for the tip!

.gitmodules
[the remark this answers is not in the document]
ferdonline: Maybe c++17 std::filesystem starts to be an option. As a header-only single-file I think it's also fine.

include/bbp/sonata/config.h
```diff
@@ -0,0 +1,149 @@
...
+
+    /**
+     * Path to the directory containing the morphologies
+     */
+    std::string morphologiesDir;
+
+    /**
+     * Dictionary for alternate directory paths.
+     */
+    std::map<std::string, std::string> alternateMorphologyFormats;
```
ferdonline, line 51: Why map and not `unordered_map` ?

src/config.cpp
```diff
@@ -0,0 +1,322 @@
...
+        components = fillComponents(json, resolver);
+        networkNodes = fillSubnetwork(networks, "node", resolver);
+        networkEdges = fillSubnetwork(networks, "edge", resolver);
+    }
+};
+
+CircuitConfig::CircuitConfig(const std::string& contents, const std::string& basePath)
+    : impl(new CircuitConfig::Impl(contents, basePath)) {}
+
+CircuitConfig::CircuitConfig(CircuitConfig&&) = default;
```
ferdonline, line 443: The compiler should generate those, none of the 5 (from rule of 5) are custom.

src/config.cpp
```diff
@@ -0,0 +1,322 @@
...
+
+template <typename Population>
+std::map<std::string, SubnetworkFiles> resolvePopulations(
+    const std::vector<SubnetworkFiles>& networkNodes) {
+    using PopulationStorage = bbp::sonata::PopulationStorage<Population>;
+    std::map<std::string, SubnetworkFiles> result;
+
+    for (const auto& network : networkNodes) {
+        const auto population = PopulationStorage(network.elements, network.types);
+        for (auto name : population.populationNames()) {
```
ferdonline, line 289: Any reason not to:
```suggestion
        const auto& population = PopulationStorage(network.elements, network.types);
        for (const auto& name : population.populationNames()) {
```

include/bbp/sonata/config.h
NadirRoGue: No reason at all, there is no need for any sort. I replaced them with `unordered_map`. Thanks!

src/config.cpp
NadirRoGue: Because the copy constructor is deleted, in the static initializer `fromFile` the compiler tries to call the copy constructor if the move constructor is not manually defined (not sure why, since as you said, it should be automatically generated). 
They also both need to be defined in the `.cpp` because `CircuitConfig::Impl` is an incomplete type in the header.

src/config.cpp
NadirRoGue: No reason at all, I replaced the values by `const &` on the for range loops. In the population statement, though, I don't understand the reference of a assigned constructed object. However, I have replace the assignment with with in-place construction to make the code cleaner.

src/config.cpp
mgeplf: > They also both need to be defined in the .cpp because CircuitConfig::Impl is an incomplete type in the header.

Yeah; good old c++.  The other option is to drop the `PIMPL`; was just matching the Brion version - up to you @NadirRoGue what you prefer.

I probably should have something like:
```
CircuitConfig& operator=(CircuitConfig&& rhs) = default;
CircuitConfig& operator=(CircuitConfig& rhs) = delete;
```
To match the `CircuitConfig::CircuitConfig(CircuitConfig&&) = default;` and deleted version.

NadirRoGue: I'd rather remove pimpl if everybody is ok with that.

ferdonline: Actually I wondered why it was used, If not too laborious, yeah, would go with that

src/config.cpp
```diff
@@ -0,0 +1,467 @@
...
+    auto populations = listNodePopulations();
+    if (populations.find(name) == populations.end())
+        throw SonataError(fmt::format("Could not find node population '{}'", name));
+
+    auto popPropertiesIt = _nodePopulationProperties.find(name);
+    if (popPropertiesIt != _nodePopulationProperties.end()) {
+        return popPropertiesIt->second;
+    }
+
+    return {"biophysical",
```
mgeplf, line 439: I don't think we can hardcode this; point neuron and projections are already being created.

src/config.cpp
```diff
@@ -0,0 +1,467 @@
...
+    auto populations = listEdgePopulations();
+    if (populations.find(name) == populations.end())
+        throw SonataError(fmt::format("Could not find edge population '{}'", name));
+
+    auto popPropertiesIt = _edgePopulationProperties.find(name);
+    if (popPropertiesIt != _edgePopulationProperties.end()) {
+        return popPropertiesIt->second;
+    }
+
+    return {"chemical_synapse",
```
mgeplf, line 455: Similar to the above; I don't think we can hardcode this.

mgeplf (commented):
After addressing the hardcoded values I flagged, and adding more tests to get the coverage up to 100%, I think we're nearly there.

src/config.cpp
NadirRoGue: Do you mean to replace it by a variable to easily change the default value, or that biophysical is not anymore the default value?

mgeplf: Sorry, I forgot that we had agreed on a default.  So, with some more tests, I think this is good to go.

src/config.cpp
```diff
@@ -0,0 +1,482 @@
...
+            throw SonataError(fmt::format("Invalid variable `{}`", name));
+        }
+    }
+
+    return variables;
+}
+
+}  // namespace
+
+class CircuitConfig::Parser
```
ferdonline, line 129: interesting design so that aux classes are not public 👍

ferdonline (approved):
LGTM

mgeplf:
LGTM

NadirRoGue (approved):

PR #144, opened 2021-05-19
update copyright

mgeplf:


README.rst
```diff
@@ -371,7 +372,7 @@ libsonata is distributed under the terms of the GNU Lesser General Public Licens
 unless noted otherwise, for example, for external dependencies.
 Refer to `COPYING.LESSER` and `COPYING` files for details.
 
-Copyright (C) 2018-2020, Blue Brain Project/EPFL and contributors.
+Copyright (c) 2018-2020 Blue Brain Project/EPFL
```
pramodk, line 375: Changing copyright would require `+2` from all contributors! 😅

pramodk (approved):
LGTM

README.rst
mgeplf: @alex4200 what is the guidance on this?

alex4200 (approved):

ASVG:
Hi Mike, 
I believe it is an joint effort with AIBS? Let me double check on the contract first. Thank you. A

mgeplf:
This is an implementation of a spec that was done completely by us, not sure what to do in that case.

README.rst
jdcourcol: from what I can see, all the past committers are part of BBP. The IP does not belong to them but BBP/EPFL.
So you can remove that "and contributors."

jdcourcol:
libsonata was not implemented jointly with AIBS.

Issue #119, opened 2021-01-15
Glob matching in NodeSets


matz-e:
Would be nice if the string matching in the NodeSet functionality (and by extension, I guess, attribute matching) could perform a glob match to not have to specify dozens of values.

(spin off from #118)

mgeplf:
I agree although I would go full regex rather than glob, personally:
so, a node_set could have:
`{"foo": {"mtype": "/[lL]5_.*"}}`

I've added the '/' as the signal that it's a regex.  To me it's the most natural marker of a regex, but that's debatable.  I'm not what the collision probability is of people using '/' as the start of a string value for attributes, but it's possible for directories...

Another option would be `{"foo": {"mtype": "regex:[lL]5_.*"}}`, or something more verbose, I suppose.

matz-e:
Fine with me, I guess. The added benefit of using full-blown regex is support by the STL for new enough standards.

With a prefix, I would suggest maybe adding `re:` and `val:` to be able to fully disambiguate (the latter being the default). Slightly shorter than `regex:` :)

tomdele:
Hi guys, 
I already have something like this in bluepysnap and bluepy. I must say, I am not really in favor of adding the "function" itself (here: `regex:`) to the argument (here : `"[lL]5_.*"`). I removed this kind of queries entirely from bluepy and bluepysnap.

I prefer to have something more generic (but more verbose) that we can apply to other kind of complex queries in the future. I am now using a ~"à la mongo" type of queries :
`{"mtype": {'$regex': '.*BP'}}`
So applied to the node_sets : 
`{'foo': {"mtype": {'$regex': '.*BP'}}}`
For the regex.

This kinde of syntax allows us to have more complex queries in the future like: 
`{"mtype": {'$complex_function': [arg1, arg2]}}`
where it is easy to catch the function name + arguments (no matter the types and the number of args).

tomdele:
I am also using this kind of syntax to create `or`, `and` and `node_set` queries. These are in a different scopes tho.

Ex : 
```
from bluepysnap import Circuit
circuit = Circuit("config_file.json")
nodes_pop = circuit.nodes["pop_name"]
nodes_pop.ids({"$node_set": 'Node12_L6_Y'})
nodes_pop.ids({"$and": [{"mtype": 'morphtype-a', "population": "default"}, {"etype": "etype-B"}]})
[...]

PR #147, opened 2021-06-07
add "power" matching

mgeplf:


include/bbp/sonata/nodes.h
```diff
@@ -53,6 +53,10 @@ class SONATA_API NodePopulation: public Population
     template <typename T>
     Selection matchAttributeValues(const std::string& attribute,
                                    const std::vector<T>& values) const;
+
+
+    // XXX: need to do this better
+    Selection regexMatch(const std::string& name, const std::string& re) const;
```
matz-e, line 59: We would never have a string attribute for edges?

src/node_sets.cpp
```diff
@@ -176,6 +176,120 @@ class NodeSetBasicNodeIds: public NodeSetRule
...
+        return fmt::format(R"("{}": {{ "{}": "{}" }})", attribute_, op2string(op_), value_);
+    }
+
+    enum class Op {
+        regex = 1,
+    };
+
+    static Op string2op(const std::string& s) {
+        if (s == "$regex") {
+            return Op::regex;
```
matz-e, line 204: So… in theory we could have a `$glob` with a small translation layer from glob→regular expression?

In the recipe we use glob matching, and it may be beneficial to offer a homogeneous interface to query stuff.

src/node_sets.cpp
```diff
@@ -176,6 +176,120 @@ class NodeSetBasicNodeIds: public NodeSetRule
...
+        }
+        throw SonataError(fmt::format("Operator '{}' not available for strings", s));
+    }
+
+    static std::string op2string(const Op op) {
+        switch (op) {
+        case Op::regex:
+            return "$regex";
+        default:
+            throw SonataError("Not Implemented NodeSetBasicOperator");
```
matz-e, line 214: Could be formatted like above

src/node_sets.cpp
```diff
@@ -176,6 +176,120 @@ class NodeSetBasicNodeIds: public NodeSetRule
...
+        if (s == "$gt") {
+            return Op::gt;
+        } else if (s == "$lt") {
+            return Op::lt;
+        } else if (s == "$gte") {
+            return Op::gte;
+        } else if (s == "$lte") {
+            return Op::lte;
+        }
+        throw SonataError(fmt::format("Operator '{}' not available for numeric", s));
```
matz-e, line 269: How about adding an operator `$range`, too?

src/nodes.cpp
```diff
@@ -8,71 +8,117 @@
...
+                                 std::string name,
+                                 UnaryPredicate pred) {
+    if (population.enumerationNames().count(name) > 0) {
+        const auto& enum_values = population.enumerationValues(name);
+        std::vector<size_t> wanted_enum_value;
+        wanted_enum_value.reserve(enum_values.size());
+
+        for (size_t i = 0; i < enum_values.size(); ++i) {
+            if (pred(enum_values[i])) {
+                wanted_enum_value.push_back(i);
```
matz-e, line 69: `wanted_enum_value` could just be a LUT, `std::vector<bool>` or this bitvector-thingie.
```suggestion
                wanted_enum_value[i] = pred(enum_values[i]);
```
Then you'd have O(1) lookup speed at negligible memory-overhead, and ditch the search later.

src/population.cpp
```diff
@@ -344,6 +345,26 @@ std::string Population::_dynamicsAttributeDataType(const std::string& name) cons
     return _getDataType(impl_->getDynamicsAttributeDataSet(name), name);
 }
 
+template <>
+Selection Population::filterAttribute(const std::string& name,
+                                      std::function<bool(const std::string)> pred) const {
+    auto dtype = impl_->getAttributeDataSet(name).getDataType();
+    if (dtype != HighFive::AtomicType<std::string>()) {
```
matz-e, line 352: Could this not be generalized to
```suggestion
    if (dtype != HighFive::AtomicType<T>()) {
```
And merged with below?

src/utils.h
```diff
@@ -10,5 +10,22 @@
...
+#include <vector>
+
+#include <bbp/sonata/population.h>
 
 std::string readFile(const std::string& path);
+
+template <typename T, class UnaryPredicate>
+bbp::sonata::Selection _getMatchingSelection(const std::vector<T>& values, UnaryPredicate pred) {
+    using bbp::sonata::Selection;
+    Selection::Values idx;
```
matz-e, line 22: `idx` short for index? `ids` as plural?

src/nodes.cpp
```diff
@@ -89,32 +135,11 @@ Selection NodePopulation::matchAttributeValues(const std::string& name, const T
...
-        std::vector<size_t> wanted_enum_value;
-        wanted_enum_value.reserve(values.size());
-        for (const auto& v : values) {
-            const auto wanted_index = std::find(enum_values.cbegin(), enum_values.cend(), v);
-            if (wanted_index != enum_values.cend()) {
-                wanted_enum_value.push_back(wanted_index - enum_values.cbegin());
-            }
-        }
+    const auto pred = [values](const std::string& v) {
+        return std::find(values.cbegin(), values.cend(), v) != values.cend();
```
matz-e, line 139: Sort + binary search?

matz-e (commented):
Alright, I think these are my initial 0.02 BTC. I've reached my mental call-graph reconstruction capacity, so I'm sorry if some of these don't make sense.

include/bbp/sonata/nodes.h
mgeplf: Currently not supported, correct.  If the need arises, we can add it later, but it's going to be an expensive query.

src/node_sets.cpp
mgeplf: I let clang-format work its magic

src/node_sets.cpp
mgeplf: Perhaps in the next one; could do it with a compound statement at the moment.

src/nodes.cpp
mgeplf: Hmm, yeah that could be interesting.

src/population.cpp
mgeplf: I need to manually specialize it, since the macros don't specialize std::string for other reasons.

src/nodes.cpp
mgeplf: done

src/utils.h
mgeplf: done

src/node_sets.cpp
matz-e: I meant with `fmt::format`, sorry, not very clear.

src/population.cpp
matz-e: Alrighty

src/nodes.cpp
```diff
@@ -89,32 +135,14 @@ Selection NodePopulation::matchAttributeValues(const std::string& name, const T
...
-            return Selection({});
-        }
+    std::vector<std::string> values_sorted(values);
+    std::sort(values_sorted.begin(), values_sorted.end());
 
-        return _getMatchingSelection<size_t>(getEnumeration<size_t>(name, selectAll()),
-                                             wanted_enum_value);
-    }
+    const auto pred = [&values_sorted](const std::string& v) {
+        return std::find(values_sorted.cbegin(), values_sorted.cend(), v) != values_sorted.cend();
```
matz-e, line 142:
```suggestion
        return std::binary_search(values_sorted.cbegin(), values_sorted.cend(), v) != values_sorted.cend();
```

src/node_sets.cpp
```diff
@@ -176,6 +176,125 @@ class NodeSetBasicNodeIds: public NodeSetRule
...
+        gte = 3,
+        lte = 4,
+    };
+
+    static Op string2op(const std::string& s) {
+        if (s == "$gt") {
+            return Op::gt;
+        } else if (s == "$lt") {
+            return Op::lt;
+        } else if (s == "$gte") {
```
matz-e, line 269: So… both BASH and FORTRAN use `.ge.` and `.le.`, maybe we should stick to `$le`/`$ge`?

python/generated/docstrings.h
```diff
@@ -75,6 +75,8 @@ Note: This does not match dynamics_params datasets)doc";
 
 static const char *__doc_bbp_sonata_NodePopulation_matchAttributeValues_2 = R"doc(Like matchAttributeValues, but for vectors of values to match)doc";
 
+static const char *__doc_bbp_sonata_NodePopulation_regexMatch = R"doc()doc";
```
matz-e, line 78: No documentation?

Also, should we explicitly mark these as extensions to the SONATA™ spec? Or is there an upstream PR amending this situation already?

include/bbp/sonata/common.h
```diff
@@ -45,5 +45,8 @@ class SONATA_API SonataError: public std::runtime_error
   public:
     explicit SonataError(const std::string& what);
 };
+
+#define THROW_IF_REACHED throw SonataError("Should never be reached");
```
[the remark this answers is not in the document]
mgeplf: it's a single liner, so afaik it's safe, has the guidance changed?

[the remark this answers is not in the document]
matz-e: only my paranoia

src/node_sets.cpp
mgeplf: I followed the discussion in #119; I guess this is a `mongodb-ism`.  I personally have no preference, but now is the time to change it.

python/generated/docstrings.h
mgeplf: thanks, need to add the doc...I will do that.

I think it's worth doing a PR upstream to the docs, at least so ppl can read about the feature

src/node_sets.cpp
matz-e: Por que no los dos?

mgeplf: syntax multiplicity for such a simple concept seems like overkill, imo

matz-e: Fair enough.

matz-e (approved):

PR #154, opened 2021-08-05
Avoid copy with ReportReader::Population

alkino:
Population is an heavy class, we should avoid copy of it.


alkino:
Hard to avoid copy with python binding

PR #158, opened 2021-09-20
Initial simulation config implementation

NadirRoGue:
First draft for the simulation config. Notes:
- I added the only the variables that are useful to find the report files and get some information form them
- I created a fake `simulation_config.json` test file by piercing together the parts found [https://bbpteam.epfl.ch/documentation/projects/Circuit%20Documentation/latest/sonata_simulation.html](here)

@mgeplf please ping whomever needs to see this


NadirRoGue:
So far the initial implementation is committed, please add your reviews. 
Im getting a 100% coverage on config locally, but I am not sure why here is failing with that http error.

mgeplf:
> Im getting a 100% coverage on config locally, but I am not sure why here is failing with that http error.

Coveralls is doing 'emergency maintenance'.

mgeplf:
I won't have time to look at this in the next couple days.  Perhaps @ferdonline or someone else from HPC (suggestions @pramodk) could have a look?

jplanasc:
@mgeplf @ferdonline @pramodk : could you please have a look? Thanks!

mgeplf:
@jplanasc > could you please have a look? Thanks!

I would love to, but I simply don't have the time - hopefully a real C++ engineer from HPC can look (and not some amateur like me), and I can do a final review.

python/tests/test.py
```diff
@@ -534,5 +534,33 @@ def test_get_population_properties(self):
...
+class TestSimulationConfig(unittest.TestCase):
+    def setUp(self):
+        self.config = SimulationConfig.from_file(
+                        os.path.join(PATH, 'config/simulation_config.json'))
+
+    def test_basic(self):
+        self.assertEqual(self.config.base_path, os.path.abspath(os.path.join(PATH, 'config')))
+
+        self.assertEqual(self.config.run.tstop, 1000)
+        self.assertTrue(self.config.run.dt - 0.025 < 0.01)
```
olupton, line 547: Should this be
```suggestion
        self.assertTrue(abs(self.config.run.dt - 0.025) < 0.01)
```
?

python/tests/test.py
```diff
@@ -534,5 +534,33 @@ def test_get_population_properties(self):
...
+
+        self.assertEqual(self.config.output.output_dir,
+                         os.path.abspath(os.path.join(PATH, 'config/output')))
+        self.assertEqual(self.config.output.spikes_file, 'out.h5')
+
+        self.assertEqual(self.config.report('soma').cells, 'Mosaic')
+        self.assertEqual(self.config.report('soma').type, 'compartment')
+        self.assertTrue(self.config.report('compartment').dt - 0.1 < 0.01)
```
olupton, line 555:
```suggestion
        self.assertTrue(abs(self.config.report('compartment').dt - 0.1) < 0.01)
```
?

src/config.cpp
```diff
@@ -124,6 +124,12 @@ Variables readVariables(const nlohmann::json& json) {
     return variables;
 }
 
+std::string toAbsolute(const fs::path& base, const std::string& pathStr) {
+    const fs::path path(pathStr);
```
olupton, line 128:
```suggestion
std::string toAbsolute(const fs::path& base, const fs::path& path) {
```
seems cleaner (and more flexible in case you want to call this with two arguments that are already paths)

src/config.cpp
```diff
@@ -478,5 +479,127 @@ const std::string& CircuitConfig::getExpandedJSON() const {
...
+            const auto debugStr = fmt::format("report {}", it.key());
+            parseMandatory(valueIt, "cells", debugStr, report.cells);
+            parseMandatory(valueIt, "type", debugStr, report.type);
+            parseMandatory(valueIt, "dt", debugStr, report.dt);
+            parseMandatory(valueIt, "start_time", debugStr, report.startTime);
+            parseMandatory(valueIt, "end_time", debugStr, report.endTime);
+            parseOptional(valueIt, "file_name", report.fileName);
+            if (report.fileName.empty())
+                report.fileName = it.key() + "_SONATA.h5";
+            else if (report.fileName.find(".h5") == std::string::npos)
```
olupton, line 553: Looks like this should be checking the string ending, not searching for `.h5` anywhere in it.
(Sadly [std::string::ends_with](https://en.cppreference.com/w/cpp/string/basic_string/ends_with) is C++20...)

src/config.cpp
```diff
@@ -478,5 +479,127 @@ const std::string& CircuitConfig::getExpandedJSON() const {
...
+
+  private:
+    const fs::path _basePath;
+    const nlohmann::json _json;
+};
+
+SimulationConfig::SimulationConfig(const std::string& content, const std::string& basePath)
+    : _jsonContent(content)
+    , _basePath(fs::absolute(basePath).lexically_normal().string()) {
+    const Parser parser(content, basePath);
```
olupton, line 569: Not sure I see the point of this `Parser` class, but I guess you have done it this way for consistency. This seems to imply a pointless copy of `content` and `basePath`. If I was starting from a clean slate I would lean towards implementing `parse{Run,Output,Reports}()` as pure functions in an anonymous namespace. This way you could call them directly in the initializer list.

src/config.cpp
```diff
@@ -478,5 +479,127 @@ const std::string& CircuitConfig::getExpandedJSON() const {
...
+
+const SimulationConfig::Run& SimulationConfig::getRun() const noexcept {
+    return _run;
+}
+
+const SimulationConfig::Output& SimulationConfig::getOutput() const noexcept {
+    return _output;
+}
+
+const SimulationConfig::Report& SimulationConfig::getReport(const std::string& name) const {
```
olupton, line 595: It's probably not worth doing anything about this now...but once C++17/20 are available then using `std::string_view` and transparent comparators might be nicer.

tests/test_config.cpp
```diff
@@ -288,3 +290,157 @@ TEST_CASE("CircuitConfig") {
         }
     }
 }
+
+TEST_CASE("SimulationConfig") {
+    SECTION("Simple") {
+        const auto config = SimulationConfig::fromFile("./data/config/simulation_config.json");
+        CHECK_NOTHROW(config.getRun());
+        CHECK(config.getRun().tstop == 1000.f);
+        CHECK(config.getRun().dt == 0.025f);
```
olupton, line 299: Conscious choice not to make these fuzzy comparisons?

olupton (commented):
Generally looks good to me, just a few minor comments/suggestions.

include/bbp/sonata/config.h
```diff
@@ -175,5 +175,106 @@ class SONATA_API CircuitConfig
...
+        std::string spikesFile;
+    };
+    // List of report parameters collected during the simulation
+    struct Report {
+        // Node sets on which to report
+        std::string cells;
+        // Report type. Possible values: "compartment", "summation", "synapse"
+        std::string type;
+        // Interval between reporting steps in milliseconds
+        float dt;
```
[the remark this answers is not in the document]
NadirRoGue: thanks! applied

python/tests/test.py
NadirRoGue: Good catch, thanks!

src/config.cpp
NadirRoGue: Agree, its better than mixing path and string arguments, thanks!

python/tests/test.py
NadirRoGue: Good catch again, thanks!

src/config.cpp
NadirRoGue: Good point, there are many weird paths that could make this fail. Ill commit a better version and ping you, thanks a lot!

src/config.cpp
NadirRoGue: Yes, its been done like that for consistency with the `CircuitConfig` class implementation. For `CircuitConfig`, the `Parser` class was created so that the parsing would be outside the `CircuitConfig` class itself, while the `Parser` class is also friend of `CircuitConfig` to allow access the private members. 

Its not necessary for `SimulationConfig`, although I also don't think its an expensive copy which usually would be performed just once, and also helps keep the code clean and more readable.

However, I don't have a strong preference for either implementation.

tests/test_config.cpp
NadirRoGue: Sorry, I don't understand the comment.
The checks are added to reach 100% of test coverage and to make sure the correct data member is filled with the correct field from the JSON file

src/config.cpp
NadirRoGue: I agree, although that would be more of a library-wide change

src/config.cpp
NadirRoGue: Done!, please have a look.

tests/test_config.cpp
matz-e: This is `catch2`, right?

Something like
```suggestion
        REQUIRE_THAT(config.getRun().tstop, WithinULP(1000.f, 1));
        REQUIRE_THAT(config.getRun().dt, WithinULP(0.025f, 1));
```
would be more generic and not rely on the numbers being tested being a power of two.

Needs
```
using Catch::Matchers::WithinULP;
```
somewhere…

NadirRoGue: Ah I understand now. Ill make the change, thanks!

include/bbp/sonata/config.h
```diff
@@ -175,5 +175,106 @@ class SONATA_API CircuitConfig
...
+ */
+class SONATA_API SimulationConfig
+{
+  public:
+    // Parameters defining global simulation settings for spike reports
+    struct Run {
+        // Biological simulation end time in milliseconds
+        float tstop;
+        // Integration step duration in milliseconds
+        float dt;
```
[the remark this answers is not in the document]
olupton: Looks like this was resolved without the patch being applied. Intentional?

include/bbp/sonata/config.h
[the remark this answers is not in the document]
olupton: Looks like this was resolved without the patch being applied. Intentional?

olupton (commented):
Looks like some changes you meant to apply were not applied?

Looks good to me once @matz-e's suggestion is applied.

NadirRoGue:
@olupton I added @matz-e suggestion, as well as the missing float initialization that somehow I missed applying though the github interface. Thanks for the review!

olupton (approved):

matz-e (approved):

src/config.cpp
```diff
@@ -478,5 +478,130 @@ const std::string& CircuitConfig::getExpandedJSON() const {
...
+
+        for (auto it = reportsIt->begin(); it != reportsIt->end(); ++it) {
+            auto& report = result[it.key()];
+            auto& valueIt = it.value();
+            const auto debugStr = fmt::format("report {}", it.key());
+            parseMandatory(valueIt, "cells", debugStr, report.cells);
+            parseMandatory(valueIt, "type", debugStr, report.type);
+            parseMandatory(valueIt, "dt", debugStr, report.dt);
+            parseMandatory(valueIt, "start_time", debugStr, report.startTime);
+            parseMandatory(valueIt, "end_time", debugStr, report.endTime);
```
ferdonline, line 548: The problem in this task is to know exactly what to do when we are missing data, what what defaults to apply. See e.g. `/gpfs/bbp.cscs.ch/project/proj12/example_data/sonata/v2/usecase3/simulation_sonata.json`. This is a minimalistic config file, in principle valid, and won't pass here.

jdcourcol:
@ferdonline can you approve this request ?
Can you follow up to use the config reader in the simulator to avoid duplication as agreed during SONATA meeting ?

include/bbp/sonata/config.h
```diff
@@ -175,5 +175,106 @@ class SONATA_API CircuitConfig
...
 
+/**
+ *  Read access to a SONATA simulation config file.
+ */
+class SONATA_API SimulationConfig
+{
+  public:
+    // Parameters defining global simulation settings for spike reports
+    struct Run {
+        // Biological simulation end time in milliseconds
```
alkino, line 186: if you write those comment as docstring compatible they will appear in documentation.

NadirRoGue: docstring compatible is in the format of
```
/**
 * Parameters defining global simulation settings for spike reports
 */
```
?

jorblancoa (approved):
LGTM!

ferdonline (dismissed):

src/config.cpp
jplanasc: I think we can resolve this conversation: mandatory fields need to be present in the files.

include/bbp/sonata/config.h
jplanasc: ping @alkino ?

alkino: yes this is valid syntax, `///` also work for single line

You can have a check after that with `.ci/check_generated_docstring.sh`

NadirRoGue: Thanks! I updated the comments so they are included

alkino (approved):

alkino:
You need to run ci/check_generated_docstring.sh and update the docstring file after that

NadirRoGue:
Thanks for the review @alkino @ferdonline @jorblancoa !
If someone could re-review it as I had to dismiss them to fix the docs, that would be great!

jorblancoa (approved):

mgeplf:
thanks @NadirRoGue!

PR #163, opened 2021-10-07
update github actions building

mgeplf:
* use ubuntu-20.04
* always build sdist and wheels, but only publish them on PyPI if a release

### DIFF
--- a/.github/workflows/publish-sdist-wheels.yml
+++ b/.github/workflows/publish-sdist-wheels.yml
@@ -1,27 +1,23 @@
-name: Publish sdist and wheels macos-manylinux
-
-on:
-  push:
-    tags:
-      - 'v[0-9]+.[0-9]+.[0-9]+'
+name: Publish sdist and wheels for macos, manylinux, and windows, publish to pypi if a release
+on: [pull_request, push]
 
 env:
-  CIBW_MANYLINUX_X86_64_IMAGE: manylinux2014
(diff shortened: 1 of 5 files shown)


.github/workflows/publish-sdist-wheels.yml
```diff
@@ -82,18 +85,21 @@ jobs:
...
     needs: [build_wheels, build_sdist]
-    name: Upload wheels to PyPI
+
     steps:
       - name: Download artifacts produced during the build_wheels and build_sdist jobs
         uses: actions/download-artifact@v2
         with:
           name: dist
-          path: dist
+          path: dist/
```
matz-e, line 102: why the `/`? `rsync`-like syntax stuff?

mgeplf: `rsync` behavior drives me crazy

in this case, I don't have a good reason - was a long time since I looked at this

matz-e (approved):

.github/workflows/publish-sdist-wheels.yml
matz-e: at least good to signify that this should be a directory…

mgeplf: uhh, yes, let's `recon` my change :)

mgeplf:
Note; tested the resulting wheels on py37 on linux, and py37 on macos 10.13

PR #164, opened 2021-10-21
Replaces binding of compartment ids for std::array<uint64_t,2>

sergiorg-hpc:
Using `py-bluepy`, we observed recently that the `ids` of a compartment report were mapped into a Python list of tuples instead of returning a pointer to the actual data in the bindings. This PR solves this issue by correctly returning an `numpy.ndarray`, which can be used automatically by `py-bluepy` without further changes.

For instance, retrieving a single row of a reference dataset that contains close to 80 million columns per timestep, takes the following:
```
Without fix: 147.01s
With fix: 13.44s
```
Any of the operations that involve the `ids` incur in large data conversions, increasing the execution times. One example is `len(view.ids)` inside `py-bluepy`, that required close to 30 seconds.

Further details of the on-going investigations are available in HPCTM-1487.

Co-authored with @jorblancoa, @matz-e, and @alkino.

### DIFF
--- a/include/bbp/sonata/common.h
+++ b/include/bbp/sonata/common.h
@@ -41,7 +41,7 @@ using NodeID = uint64_t;
 using EdgeID = uint64_t;
 using ElementID = uint32_t;
 // For performance reasons and to ease the Python bindings, the definition of
-// CompartmentID is no longer an std::pair<typeof(NodeID), typeof(ElementID)> 
+// CompartmentID is no longer an std::pair<typeof(NodeID), typeof(ElementID)>
 using CompartmentID = std::array<uint64_t, 2>;
 
 class SONATA_API SonataError: public std::runtime_error
--- a/include/bbp/sonata/common.h
+++ b/include/bbp/sonata/common.h
@@ -9,6 +9,7 @@
 
 #pragma once
 
+#include <array>
 #include <cstdint>
 #include <stdexcept>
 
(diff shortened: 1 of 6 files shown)


matz-e (dismissed):

sergiorg-hpc:
Pinging @mgeplf for approval and merge.

matz-e:
I'm just going to add some performance numbers from @sergiorg-hpc:
image

mgeplf:
FYI: I'll be looking at this in the next couple days.

include/bbp/sonata/common.h
```diff
@@ -39,6 +40,7 @@ SONATA_API const std::string version();
 using NodeID = uint64_t;
 using EdgeID = uint64_t;
 using ElementID = uint32_t;
+using CompartmentID = std::array<uint64_t, 2>;
```
mgeplf, line 43: can we add a comment here saying that this should be pairs of `<typeof(NodeID), typeof(ElementID)>`, but that we're using the extra space such that the bindings are (much!) faster?

mgeplf (commented):
Wow, this is great; thanks for the investigation!

include/bbp/sonata/common.h
sergiorg-hpc: Done!

sergiorg-hpc:
Thank you @matz-e for adding the info, and also @mgeplf for the comments and review.

If you guys agree, I'd suggest to merge this after the CI passes and create a new tag release for `libsonata`. I already have the PR in Spack if I'm not mistaken, so it should be a matter to either point to the specific commit after the merge (for temporary tags) or to the specific version tag (for formal releases).

@mgeplf, I believe you are the only one of us with permission to merge, so it is your call.

mgeplf (approved):

mgeplf:
nice work, always happy w/ perf!

PR #168, opened 2021-11-25
Add a function to get ids before dataframe

### LINKED ISSUE #149: Access report mapping without the need to load a frame
Hello,

Currently, to access the node compartment mapping, the only option available with the current report reader API is to query a frame of the report and use the ids field of the `DataFrame` object. 

It would be useful to be able to access this mapping information without having to read a full frame, as it might not be needed at the time the mapping is needed.

alkino:
Fix #149

### DIFF
--- a/include/bbp/sonata/report_reader.h
+++ b/include/bbp/sonata/report_reader.h
@@ -134,9 +134,16 @@ class SONATA_API ReportReader
         std::vector<NodeID> getNodeIds() const;
 
         /**
-         * Return the ElementIds for the passed Node
+         * Return the ElementIds for the passed Node.
+         * The return type will depend on the report reader:
+         * - For Soma report reader, the return value will be the Node ID to which the report
+         *   value belongs to.
+         * - For Element/full compartment readers, the return value will be an array with 2
+         *   elements, the first element is the Node ID and the second element is the 
+         *   compartment ID of the given Node.
          *
-         * \param node_ids limit the report to the given selection. If nullptr, all nodes are used
+         * \param node_ids limit the report to the given selection. If nullptr, all nodes in the
+         * report are used
          * \param fn lambda applied to all ranges for all node ids
          */
         typename DataFrame<KeyType>::DataType getNodeIdElementIdMapping(
--- a/include/bbp/sonata/report_reader.h
+++ b/include/bbp/sonata/report_reader.h
@@ -131,7 +131,8 @@ class SONATA_API ReportReader
         /**
          * Return selected ids.
          */
-        typename DataFrame<KeyType>::DataType getIds(const nonstd::optional<Selection>& node_ids = nonstd::nullopt) const;
+        typename DataFrame<KeyType>::DataType getIds(
+            const nonstd::optional<Selection>& node_ids = nonstd::nullopt) const;
 
         /**
          * \param node_ids limit the report to the given selection.
@@ -158,7 +159,8 @@ class SONATA_API ReportReader
         std::string time_units_;
         std::string data_units_;
         bool nodes_ids_sorted_ = false;
-        Selection::Values node_ids_from_selection(const nonstd::optional<Selection>& node_ids = nonstd::nullopt) const;
+        Selection::Values node_ids_from_selection(
+            const nonstd::optional<Selection>& node_ids = nonstd::nullopt) const;
 
         friend ReportReader;
     };
--- a/include/bbp/sonata/report_reader.h
+++ b/include/bbp/sonata/report_reader.h
@@ -133,7 +133,7 @@ class SONATA_API ReportReader
          */
         typename DataFrame<KeyType>::DataType getIds(
             const nonstd::optional<Selection>& node_ids = nonstd::nullopt,
-            std::function<void(const Range&)> fn = [](...) {}) const;
+            std::function<void(const Range&)> = nullptr) const;
 
         /**
          * \param node_ids limit the report to the given selection.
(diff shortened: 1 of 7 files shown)


NadirRoGue (dismissed):
Everything looks good

sergiorg-hpc:
@mgeplf I asked @alkino to make some changes here in regards to the `min`,`max` and `position`. Even though the code looks clear and simpler by duplicating the loop that goes through all the `ids`, this is a bit tricky, because some reports might have millions of cells and we would be wasting time by duplicating the loop and going through all the cells twice.

Thus, I suggested to try to keep the original design while still not affecting the performance of `libsonata`. A simple lambda parameter could be an option and, thus, the `fun` parameter that Nico added for illustration purposes.

src/report_reader.cpp
```diff
@@ -364,8 +365,8 @@ typename DataFrame<T>::DataType ReportReader<T>::Population::getIds(
             ids.push_back(make_key<T>(node_id, elem));
         }
 
-        // Call auxiliary function
-        fn(it->second);
+        if (fun)
```
matz-e, line 368: That's no fun… I liked `fn` better.

Braces for the `if`?

src/report_reader.cpp
```diff
@@ -279,6 +279,31 @@ std::vector<NodeID> ReportReader<T>::Population::getNodeIds() const {
     return nodes_ids_;
 }
 
+template <typename T>
+Selection::Values ReportReader<T>::Population::node_ids_from_selection(
+    const nonstd::optional<Selection>& selection) const {
+    // Simplify selection
+    // We should remove duplicates
+    // And when we can work with ranges let's sort them
+    // auto nodes_ids_ = Selection::fromValues(node_ids.flatten().sort());
```
mgeplf, line 288: I don't understand this comment

src/report_reader.cpp
```diff
@@ -314,6 +339,39 @@ std::pair<size_t, size_t> ReportReader<T>::Population::getIndex(
...
+template <typename T>
+typename DataFrame<T>::DataType ReportReader<T>::Population::getIds(
+    const nonstd::optional<Selection>& selection, std::function<void(const Range&)> fn) const {
+    typename DataFrame<T>::DataType ids{};
+
+
+    // Simplify selection
+    // We should remove duplicates
+    // And when we can work with ranges let's sort them
+    // auto nodes_ids_ = Selection::fromValues(node_ids.flatten().sort());
```
mgeplf, line 351: same with this comment

mgeplf:
I would like to get this merged before I do a libsonata release, but I'm still puzzled by the comments; can you have a look @alkino?

src/report_reader.cpp
alkino: I wrote it, but i don't remember why.
In my head it means, one interesting improvement could be to group ranges for having a single read

src/report_reader.cpp
alkino: this one should be deleted

alkino:
updated

include/bbp/sonata/report_reader.h
```diff
@@ -127,8 +127,19 @@ class SONATA_API ReportReader
...
+
+        /**
+         * Return all the node ids.
+         */
         std::vector<NodeID> getNodeIds() const;
 
+        /**
+         * Return selected ids.
+         */
+        typename DataFrame<KeyType>::DataType getIds(
```
mgeplf, line 139: should there be a python counterpart to this?

alkino: Yes, but i will avoid the std::function xD

mgeplf: Yup, makes sense :)

mgeplf: Any luck with this?

include/bbp/sonata/report_reader.h
```diff
@@ -127,8 +127,19 @@ class SONATA_API ReportReader
...
          */
         bool getSorted() const;
+
+        /**
+         * Return all the node ids.
+         */
         std::vector<NodeID> getNodeIds() const;
 
+        /**
+         * Return selected ids.
```
mgeplf, line 137: I think we should pick a more descriptive name, and describe it more accurately.  Something like `getNodeElementMapping` or `getNodeIdElementIdMapping`, thoughs?

matz-e: I'd be for `getCompartmentIds`?

It would also be useful to see what the function is used for, that should be noted.

mgeplf: not everything is a compartment, and the doc calls things [node element recordings](https://github.com/AllenInstitute/sonata/blob/master/docs/SONATA_DEVELOPER_GUIDE.md#frame-oriented-node-element-recordings), so I think we should use that terminology.

matz-e: Should that also entail renaming the `CompartmentId` datatype definition, then?

mgeplf: > renaming CompartmentId datatype definition

That's already part of the interface, so I think that ship has sailed.

matz-e: Alias & deprecate could be done already, though?

mgeplf: I will defer to the people who maintain the `libsonata` reporting parts.

mgeplf:
I changed the name of the function, and tried to make the docstrings more clear.  @NadirRoGue or @alkino could you clean them up further; things like `getIds` don't really convey what IDs are being retrieved.  In addition, the format of the return would be helpful for things like `getNodeIdElementIdMapping`

mgeplf:
Ok, I'll take it from here, I guess.  Thanks.

mgeplf (approved):

PR #169, opened 2021-11-25
Add getTimes inside SpikeReader::Population

### LINKED ISSUE #155: Query start and end time of spike reports
Hello,

Currently, the only way to know the start and end time of a spike report is by loading the whole report and search for the highest timestamp.

It would be useful to have a similar method as `ReportReader`'s `getTimes()` to directly return this value from the SpikeReader members.

alkino:
Return start and stop times.

Fix #155

### DIFF
--- a/src/report_reader.cpp
+++ b/src/report_reader.cpp
@@ -112,6 +112,10 @@ auto SpikeReader::openPopulation(const std::string& populationName) const -> con
     return populations_.at(populationName);
 }
 
+std::tuple<double, double> SpikeReader::Population::getTimes() const {
+    return std::tie(tstart_, tstop_);
+}
+
 Spikes SpikeReader::Population::get(const nonstd::optional<Selection>& node_ids,
                                     const nonstd::optional<double>& tstart,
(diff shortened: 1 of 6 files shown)


NadirRoGue (approved):
Everything looks perfect

mgeplf:
@alkino Before this goes in, can you add tests, and python bindings?

src/report_reader.cpp
```diff
@@ -112,6 +112,10 @@ auto SpikeReader::openPopulation(const std::string& populationName) const -> con
     return populations_.at(populationName);
 }
 
+std::tuple<double, double> SpikeReader::Population::getTimes() const {
```
mgeplf, line 115: can you add tests and python bindings for this?

mgeplf (changes requested):

sergiorg-hpc (approved):

src/report_reader.cpp
alkino: done

mgeplf (approved):
Thanks for the tests!

PR #179, opened 2022-02-18
Improve mem. ops. + NodeID-ElementID mapping bindings

sergiorg-hpc:
The PR improves the copy of the returned `ids` when fetching the data from the reports, as well as the Python bindings related to the NodeID-ElementID mapping (https://github.com/BlueBrain/libsonata/pull/168):

- In the former case, the `std::vector` is [swapped in a constant-time operation instead of a full copy](https://www.cplusplus.com/reference/vector/vector/swap). This is particularly useful when a large number of GIDs is requested. 
- In the latter case, the binding of the NodeID-ElementID mapping from is improved, returning a `numpy.array` instead of a list of tuples. Hence, providing directly the result generated from the C++ code without any additional copies.

Using the same ~60TB report from https://github.com/BlueBrain/libsonata/pull/174 that contains 4.2M GIDs (1.5B compartments), the following table illustrates the performance obtained by fetching a single timestep and varying the number of GIDs requested:

```
 # GIDS |            v0.1.11 |                 PR
 ------- -------------------- -------------------
      1 |  0.010879278182983 |  0.001698255538940
     10 |  5.343438386917114 |  5.512796163558960
    100 |  5.943392753601074 |  5.604890823364258
   1000 |  6.406604766845703 |  6.062043905258179
  10000 |  6.451685428619385 |  6.049137830734253
 100000 |  7.101217269897461 |  6.462664365768433
1000000 | 12.762017250061035 |  9.955524444580078
4234929 | 27.868659019470215 | 18.158316850662230
```

On the other hand, when using in Python the function `get_node_id_element_id_mapping()` from https://github.com/BlueBrain/libsonata/pull/168, the following table shows the performance obtained varying the number of GIDs requested:

```
 # GIDS |            v0.1.11 |                 PR
 ------- -------------------- -------------------
      1 |    0.0158734321594 |  0.001111268997192
     10 |    2.6325545310974 |  2.509858608245849
    100 |    2.9563860893249 |  2.782548427581787
   1000 |    3.2792141437530 |  2.988783836364746
  10000 |    5.4664847850799 |  3.832525491714477
 100000 |   24.0672097206115 |  3.506523370742798
1000000 |  225.1766004562378 |  6.213826656341553
4234929 | 1017.6884262561798 | 11.914409875869751
```

Note that there has been other attempts to improve the copies of the voltages and even the tuples for the NodeID-ElementID mapping container, but the benefits turned out to be negligible compared to the source code complexity that was required.

### DIFF
--- a/src/report_reader.cpp
+++ b/src/report_reader.cpp
@@ -421,7 +421,7 @@ DataFrame<T> ReportReader<T>::Population::get(const nonstd::optional<Selection>&
 
     // min and max offsets of the node_ids requested are calculated
     // to reduce the amount of IO that is brought to memory
-    const auto node_id_element_layout = getNodeIdElementLayout(node_ids);
+    auto node_id_element_layout = getNodeIdElementLayout(node_ids);
     const auto& node_ranges = node_id_element_layout.node_ranges;
     const auto min = node_id_element_layout.min_max_range.first;
     const auto max = node_id_element_layout.min_max_range.second;
@@ -433,11 +433,11 @@ DataFrame<T> ReportReader<T>::Population::get(const nonstd::optional<Selection>&
     // Fill times
     DataFrame<T> data_frame;
     for (size_t i = index_start; i <= index_stop; i += stride) {
-        data_frame.times.push_back(times_index_[i].second);
+        data_frame.times.emplace_back(times_index_[i].second);
     }
 
     // Fill ids
-    data_frame.ids = node_id_element_layout.ids;
+    data_frame.ids.swap(node_id_element_layout.ids);
 
     // Fill .data member
     size_t n_time_entries = ((index_stop - index_start) / stride) + 1;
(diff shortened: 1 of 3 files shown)


alkino (approved):

matz-e (approved):

src/report_reader.cpp
```diff
@@ -433,11 +433,11 @@ DataFrame<T> ReportReader<T>::Population::get(const nonstd::optional<Selection>&
     // Fill times
     DataFrame<T> data_frame;
     for (size_t i = index_start; i <= index_stop; i += stride) {
-        data_frame.times.push_back(times_index_[i].second);
+        data_frame.times.emplace_back(times_index_[i].second);
     }
 
     // Fill ids
-    data_frame.ids = node_id_element_layout.ids;
+    data_frame.ids.swap(node_id_element_layout.ids);
```
mgeplf, line 440: it's funny, I had this, but the asm turned out the same in both cases; now I'm doubting that.

mgeplf (approved):

sergiorg-hpc:
Thank you very much, @matz-e and @mgeplf!

mgeplf:
Thanks for the change

PR #182, opened 2022-03-09
Fix/assertion error

sanjayankur31:
Fixes an assertion error thrown when `_GLIBCXX_ASSERTIONS` is defined, and enables the option by default


matz-e (commented):
Seems sensible from my side, wondering if this will work on all compilers though.

sanjayankur31:
I'm not sure what the corresponding flags for clang are, but I see that the CI builds completed fine so perhaps these extra definitions are simply ignored?

src/report_reader.cpp
```diff
@@ -470,7 +470,7 @@ DataFrame<T> ReportReader<T>::Population::get(const nonstd::optional<Selection>&
                 data_ptr[offset] = buffer[gid_start];
             } else {  // Elements report
                 uint64_t gid_end = range.second - min;
-                std::copy(&buffer[gid_start], &buffer[gid_end], &data_ptr[offset]);
+                std::copy(buffer.begin() + gid_start, buffer.begin() + gid_end, &data_ptr[offset]);
```
sergiorg-hpc, line 473: Hmmmm... Do you mind to clarify why the change? I don't mind, but if we do change this, we should also change the code in line 470 and we must verify that the performance of the copies remains the same (e.g., this operation is executed thousands of times in thousands of timesteps, it matters!).

On a side note, perhaps using something like `std::next(buffer.begin(), gid_start)` as well.

mgeplf: Turning on the `_GLIBCXX_ASSERTIONS` makes doing the subscript on buffer look like `&buffer[buffer.size()]`, which would be an out of bounds error and thus undefined behavior.  AFAIK, this is ok in this *specific* case as long as `std::copy` doesn't deref the pointer.

The change means we use the `std::copy` overload that takes iterators, and is "safer", because there is no chance there is an OOB deref, as long as `gid_end` never is larger than the size of the array.

The ASM will look the same from both of them, I'd expect, at a reasonable optimization level.

CMakeLists.txt
```diff
@@ -49,6 +49,8 @@ set(CMAKE_CXX_STANDARD_REQUIRED ON)
 
 set(SONATA_INCLUDE_DIR ${PROJECT_SOURCE_DIR}/include)
 
+set(SONATA_COMPILE_OPTIONS "${SONATA_COMPILE_OPTIONS}" "-Wp,-D_GLIBCXX_ASSERTIONS")
```
mgeplf, line 52: not a huge fan of using glibcxx specific options (ie: things that libc++ doesn't support).

However, I do like catching these sorts of things, so I'm torn.  At the minimum, I think we should only enable it for `SONATA_CXX_WARNINGS`, and gate it by the c++ library being used, which I will have to look into.

We'll take it from here; thanks @sanjayankur31 very much, this is excellent feedback.

sanjayankur31: Thanks very much @mgeplf . We'll have libsonata included in Fedora soon and then it'll be built for all the architectures we support. I'll report anything else that we catch too with PRs where possible.

src/report_reader.cpp
sergiorg-hpc: Thanks @mgeplf for clarifying. My point was that accessing the array with `[]` in `std::copy` might avoid boundary checks or additional code, but if you feel that it should be the same, then no complains at all.

I'd still prefer the use of `std::next(buffer.begin(), gid_start)` and `std::next(buffer.begin(), gid_end)` with iterators, although I don't mind leaving the code as it is with the `+` operator.

mgeplf: > but if you feel that it should be the same, then no complains at all.

Had a quick look on godbolt to confirm (which I never take as conclusive), but it seems that compiler sees through this:
https://godbolt.org/z/8dez5nEM3

> I'd still prefer the use of std::next(buffer.begin(), gid_start)

Ok, I'll make those changes, and another one to remove the pointer usage altogether.

mgeplf: I've made the changes; any/all reviews welcome.

sanjayankur31: All tests pass in my Fedora build with the newest changes :+1:

sergiorg-hpc (approved):
Sorry for the delay, looks good to me :+1:. Thank you @mgeplf and @sanjayankur31 for the improvements!

PR #188, opened 2022-03-31
Add a check when feeding a directory path to readFile

NadirRoGue:
While trying to read a SONATA config file, by mistake I generated a path that pointed to the folder contained the config file, rather than the file itself.

`file.fail()` evaluated to false, so it tried to read the directory.

`file.tellg()` returned 0, so it tried to allocate 0 bytes for the std::string which resulted on a segfault

### DIFF
--- a/src/utils.cpp
+++ b/src/utils.cpp
@@ -14,7 +14,6 @@
 #include <fstream>
 
 std::string readFile(const std::string& path) {
-
     namespace fs = ghc::filesystem;
 
     if (!fs::is_regular_file(path))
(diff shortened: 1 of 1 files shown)


src/utils.cpp
```diff
@@ -9,10 +9,16 @@
 
 #include "utils.h"
 
-#include <fstream>
+#include "../extlib/filesystem.hpp"
 
+#include <fstream>
 
 std::string readFile(const std::string& path) {
+    namespace fs = ghc::filesystem;
```
mgeplf, line 17: nice, thanks

mgeplf (approved):

PR #192, opened 2022-04-06
New parameters in SimulationConfig

WeinaJi:
- Parse floating point with type "double" to align with parser "nlohmann::json"
- Parse variables in "manifest" section and apply them to the configuration file
- Parse "network" parameter for the path of circuit_config file

### DIFF
--- a/include/bbp/sonata/config.h
+++ b/include/bbp/sonata/config.h
@@ -186,9 +186,9 @@ class SONATA_API SimulationConfig
      */
     struct Run {
         /// Biological simulation end time in milliseconds
-        float tstop{};
+        double tstop{};
         /// Integration step duration in milliseconds
-        float dt{};
+        double dt{};
     };
     /**
      * Parameters to override simulator output for spike reports
@@ -208,11 +208,11 @@ class SONATA_API SimulationConfig
         /// Report type. Possible values: "compartment", "summation", "synapse"
         std::string type;
         /// Interval between reporting steps in milliseconds
-        float dt{};
+        double dt{};
         /// Time to step reporting in milliseconds
-        float startTime{};
+        double startTime{};
         /// Time to stop reporting in milliseconds
-        float endTime{};
+        double endTime{};
         /// Report filename. Default is "<report name>_SONATA.h5"
         std::string fileName;
     };
@@ -265,6 +265,8 @@ class SONATA_API SimulationConfig
      */
     const Report& getReport(const std::string& name) const;
 
+    const std::string& getNetwork() const noexcept;
+
   private:
     // JSON string
     const std::string _jsonContent;
@@ -277,6 +279,8 @@ class SONATA_API SimulationConfig
     Output _output;
     // List of reports
     std::unordered_map<std::string, Report> _reports;
+    // Path of circuit config file for the simulation
+    std::string _network;
 
     class Parser;
     friend class Parser;
--- a/src/config.cpp
+++ b/src/config.cpp
@@ -100,9 +100,15 @@ nlohmann::json expandVariables(const nlohmann::json& json,
 
 using Variables = std::map<std::string, std::string>;
 
-// parse variables named like $[a-zA-Z0-9_]* like in manifest sections
-Variables parseManifest(const nlohmann::json& manifest) {
+Variables readVariables(const nlohmann::json& json) {
     Variables variables;
+
+    if (json.find("manifest") == json.end()) {
+        return variables;
+    }
+
+    const auto manifest = json["manifest"];
+
     const std::regex regexVariable(R"(\$[a-zA-Z0-9_]*)");
 
     for (auto it = manifest.begin(); it != manifest.end(); ++it) {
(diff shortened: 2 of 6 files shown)


include/bbp/sonata/config.h
```diff
@@ -208,11 +208,11 @@ class SONATA_API SimulationConfig
         /// Report type. Possible values: "compartment", "summation", "synapse"
         std::string type;
         /// Interval between reporting steps in milliseconds
-        float dt{};
+        double dt{};
```
mgeplf, line 211: `dt`/`startTime`/`endTime` are all floats in the spec for the report, so I think it makes sense to keep the same type:

https://github.com/AllenInstitute/sonata/blob/master/docs/SONATA_DEVELOPER_GUIDE.md#simulation-output---reports

thoughts @NadirRoGue?

NadirRoGue: Unless precision is a problem, I prefer floats as well

WeinaJi: Yes, precision is a problem for neurodamus. While reading out the floating values from libsonata, such as `dt`, the precision is different from the original value in the json file, which leads to different simulation results. As python uses double precision, as well as the json parser https://json.nlohmann.me/api/basic_json/number_float_t/ . I think it is better to align.

src/config.cpp
```diff
@@ -124,6 +118,16 @@ Variables readVariables(const nlohmann::json& json) {
     return variables;
 }
 
+// read variables in manifest section for circuit_config
+Variables readVariables(const nlohmann::json& json) {
+    Variables variables;
+
+    if (json.find("networks") == json.end() || json.find("manifest") == json.end()) {
```
WeinaJi, line 125: @mgeplf : is the check of `networks` section mandatory for CircuitConfig while parsing manifest? Otherwise I could reuse the function`readVariables` for SimulationConfig without splitting it.

include/bbp/sonata/config.h
mgeplf: > which leads to different simulation results

How does it lead to different results?
From the spec for the reports, the values are in floats, so how is that conversion taking place safely?

src/config.cpp
mgeplf: Yes, it's mandatory according to the specification:
https://bbpteam.epfl.ch/documentation/projects/circuit-documentation/latest/sonata_config.html#networks

mgeplf: However, if that's being validated somewhere else, we can rely on that.  We may need to add more test cases to make sure we're not missing anything if we change this.

mgeplf: Looks like it's validated ~line 176, so it should be safe to remove from `readVariables`

include/bbp/sonata/config.h
WeinaJi: Hello @mgeplf (cc @pramodk @ferdonline ) , in python and also py-neurodamus when I read a simulation config file containing `"dt": 0.025` via libsonata (master branch), I get this value:
```
>>> from libsonata import SimulationConfig
>>> conf = SimulationConfig.from_file("simulation_config.json")
>>> conf.run.dt
0.02500000037252903
>>> print("%.15f" %conf.run.dt)
0.025000000372529
```
neurodamus passes it to NEURON which also uses double type for floating, and the result is different (tested by CI). I see that the SONATA spec uses `float` type for all time variables, and on the other side the parser `nlohmann::json` uses `double`.  Although in python I could correct that value by `numpy.float32(dt)`, I think it would be better that libsonata outputs the right value. What do you think?

BTW, maybe it would work by changing the default type to `float` in the parser, as mentioned [here](https://github.com/nlohmann/json/issues/1109#issuecomment-392255532). But it didn't work for me, still trying.

mgeplf:
```
>>> print("%.15f" %conf.run.dt)
0.025000000372529
```
comes from printing `%.15f`, if `%.17f' is used, you get the same as the first case: 0.02500000037252903

Doubles (float64) are a strict superset of float32, so converting from float32 -> to float64 should always give the same representation.  Once it's been in float32, it should always be the same:
Ex: https://godbolt.org/z/fzMGz17Wx

That means that if the float64 parsed from nlohmann gets cast to float32 in libsonata, and should be the same in all cases, whether it's read as a float32 or float64.

If the simulation you're comparing in to in the CI was using a different a different parser to get the dt, it might always have been in float64, and thus the number will be different.  That's why I wrote above that keeping the same type as the expected report output makes sense: otherwise that conversion will happen at some other point, and it won't be easy/possible to reconcile what was written to the report versus what was simulated.

jorblancoa: Actually the report saves this info as double as per spec:
https://github.com/AllenInstitute/sonata/blob/master/docs/SONATA_DEVELOPER_GUIDE.md#output-file-formats

So it will be converted to double for the creation of the report anyway...

ferdonline: float->double is not a problem, but reading as float is the step where we lose some decimal places and enter different results land.
Being very objective, using doubles will not make a difference to anyone except ourselves who are bound to ensure reproducibility. And this being the case I think it makes sense to change to double to ensure a smooth transition.
Even if someone really wants a float, well, truncating is always possible, just the opposite is not.

mgeplf: @jorblancoa > Actually the report saves this info as double as per spec: 

Yup, that's the important one - I missed that:
`/report/<population_name>/mapping/time (dtype: double, shape: 3): the values of the data set are start time, stop time and time step`

`double` it is.

mgeplf:
Looks reasonable to me, if @NadirRoGue is good, we'll merge it.

NadirRoGue:
Looks good to me as well, thanks!

mgeplf (approved):
Thanks!

WeinaJi:
Thanks, @mgeplf  @NadirRoGue . Can you merge this PR ?  I don't have the permission.

mgeplf:
Done, thanks.  I will make a release in a bit

PR #194, opened 2022-04-21
Parsing report SimulationConfig fields

jorblancoa:


src/config.cpp
```diff
@@ -546,11 +546,55 @@ class SimulationConfig::Parser
...
             parseMandatory(valueIt, "dt", debugStr, report.dt);
             parseMandatory(valueIt, "start_time", debugStr, report.startTime);
             parseMandatory(valueIt, "end_time", debugStr, report.endTime);
             parseOptional(valueIt, "file_name", report.fileName);
+            parseOptional(valueIt, "enabled", report.enabled);
+
+            if (report.sections.empty()) {
+                report.sections = "soma";
+            } else {
+                std::set<std::string> possibleValues = {"soma", "axon", "dend", "apic", "all"};
```
mgeplf, line 564: can we add a helper function to do this test?  I think it would be nice to be able to do it with having to create a `std::set`, too.

src/config.cpp
```diff
@@ -546,11 +546,55 @@ class SimulationConfig::Parser
...
+                                                  fmt::join(possibleValues, ", ")));
+                }
+            }
+
+            if (report.scaling.empty()) {
+                report.scaling = "area";
+            } else {
+                std::set<std::string> possibleValues = {"none", "area"};
+                if (possibleValues.find(report.scaling) == possibleValues.end()) {
+                    throw SonataError(fmt::format("Scaling '{}' not supported ('{}') possible",
```
mgeplf, line 577: i think the failing case is missing a test at the moment

mgeplf (commented):
thanks! looks good, just a couple notes

src/config.cpp
```diff
@@ -546,11 +546,55 @@ class SimulationConfig::Parser
...
+
+            if (report.sections.empty()) {
+                report.sections = "soma";
+            } else {
+                std::set<std::string> possibleValues = {"soma", "axon", "dend", "apic", "all"};
+                if (possibleValues.find(report.sections) == possibleValues.end()) {
+                    throw SonataError(fmt::format("Section '{}' not supported ('{}') possible",
+                                                  report.sections,
+                                                  fmt::join(possibleValues, ", ")));
+                }
```
WeinaJi, line 569: Maybe we can have a function for this check ? It will be used in many places I guess.

src/config.cpp
mgeplf: ahem, that's *without* creating a std::set.

mgeplf: an `initializer_list` probably works:
```
template<typename T>
bool isin(const T& needle, std::initializer_list<T>haystack)
{
    return std::find(std::begin(haystack),
                     std::end(haystack),
                     needle) != std::end(haystack);
}
```

jorblancoa: I tried but I was getting errors.
Apparently it doesn't work with strings:
https://www.geeksforgeeks.org/efficient-ways-to-compare-a-variable-with-multiple-values/

Its a pity we don't have C++17 since there is a more elegant solution...
```
template <typename First, typename... T>
bool is_in(First&& first, T&&... t)
{
    return ((first == t) || ...);
}
string c = "abc";
// Compare using template
if (is_in(c, "xyz", "bhy", "abc"))
```

jorblancoa: actually with initializer_list it would work but you need to cast it
```
std::string value = "val1";
if (is_in(val, std::initializer_list<std::string>{ "val1", "val2", "val3" })) {
...
```

mgeplf: doesn't it work by giving the template param:

ie: `if (is_in(c, "xyz", "bhy", "abc"))` ->`if (is_in<std::string>(c, "xyz", "bhy", "abc"))`

pramodk:
@mgeplf: could you take a look at this again so that we can get this in?

src/config.cpp
```diff
@@ -546,11 +560,45 @@ class SimulationConfig::Parser
             auto& valueIt = it.value();
             const auto debugStr = fmt::format("report {}", it.key());
             parseMandatory(valueIt, "cells", debugStr, report.cells);
+            parseOptional(valueIt, "sections", report.sections);
             parseMandatory(valueIt, "type", debugStr, report.type);
+            parseOptional(valueIt, "scaling", report.scaling);
+            parseOptional(valueIt, "compartments", report.compartments);
+            parseMandatory(valueIt, "variable_name", debugStr, report.variableName);
```
mgeplf, line 567: for `variable_name`, I see the spec says there is the possibility of comma separated values; it should probably be checked for correctness (ie: trailing `,`, double comma, etc)

include/bbp/sonata/config.h
```diff
@@ -205,8 +205,21 @@ class SONATA_API SimulationConfig
...
         /// Node sets on which to report
         std::string cells;
+        /// Sections on which to report.
+        /// Possible values: "soma"(default), "axon", "dend", "apic", "all"
+        std::string sections;
         /// Report type. Possible values: "compartment", "summation", "synapse"
         std::string type;
+        /// For summation type, specify the handling of density values.
+        /// Possible values: "none", "area"(default)
+        std::string scaling;
```
mgeplf, line 215: Should we have everything that looks like an `enum` be an enum, rather than using strings?  I'm used to stringly typed things from python, but in c++ I think the style leans towards stronger typing.

@jorblancoa I see you were following the style of `type` above, but this is a good chance to change it if we want to.

@NadirRoGue do you have a feeling about this?

jorblancoa: My only concern is that since Neurodamus needs a string at the end, we will need to convert the enum to the string either in libsonata or later in neurodamus...

mgeplf: > My only concern is that since Neurodamus needs a string at the end,

Can you point me to where neurodamus would need a string?  The enums would be exposed, so being able to access them and compare that way should work.

NadirRoGue: If feasible, the enum would be less prone to mistakes and IDE-friendly

jorblancoa: Of course is possible, but the code would need to be changed in several places in Neurodamus/CoreNEURON
few examples:
https://bbpgitlab.epfl.ch/hpc/sim/neurodamus-py/-/blob/main/neurodamus/node.py#L884
https://bbpgitlab.epfl.ch/hpc/sim/neurodamus-py/-/blob/main/neurodamus/target_manager.py#L464
https://bbpgitlab.epfl.ch/hpc/sim/neurodamus-core/-/blob/main/hoc/Report.hoc#L86

mgeplf: > If feasible, the enum would be less prone to mistakes and IDE-friendly, IMO

I agree; the advances in warning if `switch()` statements contain all cases is really handy.

> changed in several places in Neurodamus/CoreNEURON few examples:

Thanks for the examples.  It looks like updating the python side would be ok, I think.  The HOC side of things would be a bit harder; can the enum be exported from python, so that the correspondence between enum value and representation stays the same?

jorblancoa: >can the enum be exported from python, so that the correspondence between enum value and representation stays the same?

We could pass an integer (the enum position) to hoc and from there parsing assuming we know which number corresponds to which type.

Another thing that I didn't mention is that we write from a mod file the type of the report as a string to the report.conf file , and later this is read from CoreNEURON.

Also we would need to translate in libsonata the string read from the sonata_config.json to the correspondant enum, while now it just gets asigned to the string variable.
https://github.com/BlueBrain/libsonata/blob/80fae0845a2fa8675e2db91232706c8bfa4aef27/src/config.cpp#L514
https://github.com/BlueBrain/libsonata/blob/80fae0845a2fa8675e2db91232706c8bfa4aef27/src/config.cpp#L521

mgeplf: > We could pass an integer (the enum position) to hoc and from there parsing assuming we know which number corresponds to which type.

I was hoping there was something we could do that's more elegant - that seems error prone; is there a way we could put something in the hoc scope that would automatically have these values populated?

> and later this is read from CoreNEURON.

iirc, `CoreNEURON` uses `libsonata`, so it would be able to take advantage of the the enums that were created there, so it would be safe to serialize only the enum int value.

> Also we would need to translate in libsonata the string read [...]

Yeah, one has to write a `toString`/`fromString`; it's annoying boiler plate, but it's pretty much the standard for dealing with enums in C++ afaik.

---

The tradeoff I'm seeing is that we have to balance the current convenience of using only `string` with getting the safety of having enums, and having to do some refactoring in other locations.  The latter could bring additional benefits, but those are hard to quantify at the moment.  Am I reading that right?

jorblancoa: Yes exactly.
I created a draft PR also with some proof of concept on how it will look with enums
https://github.com/BlueBrain/libsonata/pull/197

I think is quite a bit of work in some places (specially hoc as we discussed) on top of having quite a bit extra lines of code aswell in libsonata.

mgeplf: I think the #197 looks like the right way to go - NLOHMANN_JSON_SERIALIZE_ENUM is a great find, didn't know about it.

I don't mind the extra lines in libsonata, especially if it gains us correctness and maintainability.

IIRC, `pybind11` gives us `__str__` for enums, so the initial burden of change over should be ok.

If it's ok with you, I'll take that PR, and augment it.  In the meantime, I think we should be close to getting this in.

src/config.cpp
```diff
@@ -546,17 +567,39 @@ class SimulationConfig::Parser
...
+                                       report.fileName,
+                                       it.key() + "_SONATA.h5");
+            parseOptional(valueIt, "enabled", report.enabled);
+
+            checkValidField<std::string>(report.sections, {"soma", "axon", "dend", "apic", "all"});
+            checkValidField<std::string>(report.scaling, {"none", "area"});
+            checkValidField<std::string>(report.compartments, {"center", "all"});
+
+            // Validate comma separated strings
+            std::regex expr("^\\w+(\\s*,\\s*\\w+)*$");
```
mgeplf, line 594: IIRC, the `^` and `$` aren't needed for `regex_match`, since it matches the whole thing.

The grouping can be non-capturing, since we don't extract results.

Using a raw string literal will also mean the double backslashes aren't needed.

jorblancoa: Lets close then this PR and continue the conversation in #197

include/bbp/sonata/config.h
jorblancoa: I was thinking about this, however by default if you convert the enum to string you will get:
```
Type.compartment
Sections.soma
...
```
I was trying to overload the __str__ in the bindings to return just "compartment" or "soma", but is not as easy at it seems
https://github.com/pybind/pybind11/issues/2537

mgeplf: the enum value `.name` should work, no?

PR #195, opened 2022-04-27
Updating copyright year 2022

bbpgithubaudit:


mgeplf (approved):

pramodk:
@mgeplf : how you are planning to address/fix other copyright year mentioned in the rest of the source: https://github.com/BlueBrain/libsonata/blob/da8291ac4694e2e74cba68adb67bda5e652b65c3/src/hdf5_mutex.cpp#L2

I am seeing the same situation about PR's created via @bbpgithubaudit and hence asking.

mgeplf:
> @mgeplf : how you are planning to address/fix other copyright year mentioned in the rest of the source:
> I am seeing the same situation about PR's created via @bbpgithubaudit and hence asking.

Good question; in other OSS projects, ppl seem to update the copyright date when the file is updated, so maybe that's a way to do it?

matz-e:
Spack does bulk updates to the copyright, just one commit / year.

pramodk:
> ppl seem to update the copyright date when the file is updated, so maybe that's a way to do it?

This is resulting in a bit of inconsistency I think.  Something like Matthias mentioned might be better? 

By the way, copyright years mentioned in the source repos are also different/inconsistent. Maybe @bbpgithubaudit can help to `align` that and such updates could be done in a more robust + automated way.

mgeplf:
> This is resulting in a bit of inconsistency I think.

Yeah, just noting that how other places do it.

> Maybe @bbpgithubaudit can help to align that and such updates could be done in a more robust + automated way.

Yup, that would be good.  Not sure if @alex4200 gets notified w/ @bbpgithubaudit , but I'm looping him in now.

mgeplf:
I've mentioned the `SPDX` thing before, and that way the headers are small, and the date isn't required:
ex: https://github.com/torvalds/linux/blob/master/arch/arc/boot/dts/Makefile

PR #197, opened 2022-05-05
Use enum while parsing certaing string fields

jorblancoa:


src/config.cpp
```diff
@@ -546,17 +595,51 @@ class SimulationConfig::Parser
...
+                                       report.fileName,
+                                       it.key() + "_SONATA.h5");
+            parseOptional(valueIt, "enabled", report.enabled);
+
+            checkValidEnum(report.sections);
+            checkValidEnum(report.scaling);
+            checkValidEnum(report.compartments);
+
+            // Validate comma separated strings
+            std::regex expr(R"(\w+(\s*,\s*\w+)*)");
```
mgeplf, line 634: are the capture groups needed?

python/bindings.cpp
```diff
@@ -537,11 +537,29 @@ PYBIND11_MODULE(_libsonata, m) {
...
                       &SimulationConfig::Report::type,
                       "Report type. Possible values are 'compartment', 'summation', 'synapse")
+        .def_readonly("scaling",
+                      &SimulationConfig::Report::scaling,
+                      "For summation type, specify the handling of density values. "
+                      "Possible values are 'none', 'area'")
+        .def_readonly("compartments",
+                      &SimulationConfig::Report::compartments,
+                      "For compartment type, select compartments to report. "
+                      "Possible values are 'center', 'all'")
```
mgeplf, line 556: don't need possible values imo; the enum specifies that; same w/ above

include/bbp/sonata/config.h
```diff
@@ -203,10 +203,28 @@ class SONATA_API SimulationConfig
...
     struct Report {
+        enum class Sections { invalid = -1, soma, axon, dend, apic, all };
+        enum class Type { invalid = -1, compartment, summation, synapse };
+        enum class Scaling { invalid = -1, none, area };
+        enum class Compartments { invalid = -1, center, all };
+
         /// Node sets on which to report
         std::string cells;
+        /// Sections on which to report.
+        /// Possible values: "soma"(default), "axon", "dend", "apic", "all"
```
mgeplf, line 214: Only need to say what the defaults are, not the possible values; the enum describes that

NadirRoGue:
LGTM

src/config.cpp
jorblancoa: Yes, otherwise it doesn't compile.
https://www.geeksforgeeks.org/raw-string-literal-c/

mgeplf: The inner one though: `(\s*,\s*\w+)`?

jorblancoa: Then single values fail ("variable") and wrong comma separation pass ("variable,")

mgeplf: But they're not being captured, correct, it's just to group.  Ie:
`\w+(?:\s*,\s*\w+)*` would work?

jorblancoa: do you mean this?
`std::regex expr(R"(\w+(?:\s*,\s*\w+)*)");`

mgeplf: yeah; I think so

python/tests/test.py
```diff
@@ -592,12 +593,20 @@ def test_basic(self):
         self.assertEqual(self.config.output.spikes_file, 'out.h5')
 
         self.assertEqual(self.config.report('soma').cells, 'Mosaic')
-        self.assertEqual(self.config.report('soma').type, 'compartment')
+        self.assertEqual(self.config.report('soma').type, Report.Type.compartment)
+        self.assertEqual(self.config.report('soma').compartments, Report.Compartments.center)
```
WeinaJi, line 597: I am wondering how we are going to use the enum values in neurodamus?
Use the enum type directly
`self.assertEqual(self.config.report('soma').type, Report.Type.compartment)`
or still convert to string?
`self.assertEqual(self.config.report('soma').type.name, 'compartment')`

jorblancoa: I created a neurodamus-py MR for that
https://bbpgitlab.epfl.ch/hpc/sim/neurodamus-py/-/merge_requests/47

WeinaJi: Good, I see.  Can you also give an example in the python test here using `.name` ?

WeinaJi:
Thanks! Looks all good to me.

mgeplf:
Yup, likewise; thanks!

I'll merge it after the branch update.

jorblancoa:
Now that i was double checking, the `checkValidField` and `is_in` methods are not used anymore...

src/config.cpp
```diff
@@ -129,6 +151,27 @@ std::string toAbsolute(const fs::path& base, const fs::path& path) {
...
+void checkValidField(const Type& field, const std::initializer_list<Type>& possibleValues) {
+    if (!is_in<Type>(field, possibleValues)) {
+        throw SonataError(fmt::format("Field '{}' not supported ('{}') possible",
+                                      field,
+                                      fmt::join(possibleValues, ", ")));
+    }
+}
+
+template <typename Type>
+void checkValidEnum(const Type& field) {
```
jorblancoa, line 169: Shall I add something here so the error message is more descriptive? (i guess we would need an enum.toString...)

WeinaJi: Yes, sounds better, message like in `checkValidField`. In fact, I just notice, where is `checkValidField` used ?

jorblancoa: @WeinaJi i noticed also that they are not used and will remove them.
@mgeplf do you think is worth add a string member for the enum fields in order to get a better error message in checkValidEnum() method?
```
Type type;
string type_name;
...
```

mgeplf: > do you think is worth add a string member for the enum fields in order to get a better error message in checkValidEnum() method?

not sure what you mean by a `string member`?

Having a hint about which key is being checked would be a minimum, in my opinion.

One option would be to logic to `parseMandatory` / `parseOptional` that `std::enable_if` by `std::is_enum` that adds better error messages?  That way, the erroneous value is easily accessible, and it can be included in the error?

jorblancoa: Nevermind, I think printing the key is enough, I feel adding the string value in the error message is too complicated.
You can go ahead and merge whenever you feel!

mgeplf:
Thanks! This is a big step forward; appreciate the work.

PR #201, opened 2022-05-18
use doc handling to avoid repeating the same docstring in python bindings

mgeplf:


mgeplf:
> Btw, how do we generate docstrings.h ? I am trying on bb5 but I get

I should write this up more clearly; in short, the CI is setup to use `clang-9`'s libclang, because different versions give (slightly) different results.

If you have an ubuntu-20.04 desktop, you can use the `libclang1-9` package, and run the tool like:
```
LIBCLANG_PATH=/usr/lib/x86_64-linux-gnu/libclang-9.so.1 ./ci/check_generated_docstrings.sh
```

If you have issues with that, don't spend too much time on it, and contact me on slack and I'll help.

PR #202, opened 2022-05-23
Use Variants for Simulation::Input module parameters

mgeplf:
Each SimulationConfig input has a module, each with their own parameters. To reduce the chance of using the wrong parameters, we can use a variant, and return a struct w/ only the parameters that are useful for that particular module.
Ex: `Subthreshold` only has `percentLess`, instead of the union of all parameters for all modules.

### DIFF
--- a/src/config.cpp
+++ b/src/config.cpp
@@ -252,8 +252,6 @@ SimulationConfig::Input parseInputModule(const nlohmann::json& valueIt,
                                          const std::string& debugStr) {
     using Module = SimulationConfig::InputBase::Module;
 
-    const auto moduledebugStr = fmt::format("Unknown module for the input_type in {}",
-                                            debugStr);
     const auto parseCommon = [&](auto& input) {
         input.module = module;
         parseMandatory(valueIt, "input_type", debugStr, input.inputType);
@@ -307,12 +305,10 @@ SimulationConfig::Input parseInputModule(const nlohmann::json& valueIt,
         }
 
         if (ret.mean.has_value() && ret.meanPercent.has_value()) {
-            throw SonataError(
-                fmt::format("Both `mean` or `mean_percent` have values in {}", debugStr));
+            throw SonataError("Both `mean` or `mean_percent` have values in " + debugStr);
         } else if (!ret.mean.has_value() && !ret.meanPercent.has_value()) {
-            throw SonataError(
-                fmt::format("One of `mean` or `mean_percent` need to have a value in {}",
-                            debugStr));
+            throw SonataError("One of `mean` or `mean_percent` need to have a value in " +
+                              debugStr);
         }
 
         parseOptional(valueIt, "variance", ret.variance);
@@ -363,7 +359,7 @@ SimulationConfig::Input parseInputModule(const nlohmann::json& valueIt,
         return ret;
     }
     default:
-        throw SonataError(moduledebugStr);
+        throw SonataError("Unknown module for the input_type in " + debugStr);
     }
 }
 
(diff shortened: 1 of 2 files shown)


python/tests/test.py
```diff
@@ -615,24 +615,30 @@ def test_basic(self):
...
-        self.assertEqual(self.config.input('ex_pulse').width, 1)
-        self.assertEqual(self.config.input('ex_pulse').frequency, 80)
-        self.assertEqual(self.config.input('ex_noise_meanpercent').mean_percent, 0.01)
-        self.assertEqual(self.config.input('ex_noise_meanpercent').mean, None)
-        self.assertEqual(self.config.input('ex_noise_mean').mean, 0)
-        self.assertEqual(self.config.input('ex_noise_mean').mean_percent, None)
-        self.assertEqual(self.config.input('ex_rel_shotnoise').random_seed, self.config.run.random_seed)
-        self.assertEqual(self.config.input('ex_rel_shotnoise').dt, 0.25)
-        self.assertEqual(self.config.input('ex_replay').spike_file,
+        self.assertEqual(self.config.input('ex_linear').parameters.amp_start, 0.15)
```
WeinaJi, line 621: This PR is a nice improvement.  I have just one concern:
In neurodamus, currently we retrieve all the parsed values from libsonata via function like
`getattr(libsonata_obj, sonata_key)`
This PR changes the API a bit. For certain parameters of an input object, I need to adapt neurodamus to call `getattr(libsonata_obj.parameters, amp_start)`. Is it possible to avoid the layer `parameters`?
cc @ferdonline

mgeplf: Yeah, that could be done.  Right now it's clear what is common to all stimulus (`module/inputType/delay/duration/nodeSet`) and what is unique to each.

> getattr(libsonata_obj.parameters, amp_start)

Is there a reason why this can't be `libsonata_obj.parameters.amp_start`?

WeinaJi: >Is there a reason why this can't be libsonata_obj.parameters.amp_start?

It is to avoid additional coding in neurodamus to adapt this change. It would be great if we can call `libsonata_obj.<key_name>` for both common and unique keys. Please let me know if it is feasible. Thanks!

mgeplf: It's feasible, from an ergonomic point of view it's less clean, especially on the C++ side, I would say - one has to either use `visit` because one can't look at the value of the `module` before picking the correct variant.

I'll do and push the change, and we can decide then.

mgeplf: > I'll do and push the change, and we can decide then.

I did the change, let me know what you think.

WeinaJi: Yes, it looks like what I was asking for. I will review it today.

src/config.cpp
```diff
@@ -717,102 +840,41 @@ class SimulationConfig::Parser
...
+
+            const auto input =
+                parseInputModule(valueIt, module, _basePath, parseRun().randomSeed, debugStr);
+            result[it.key()] = input;
+
+            auto inputType = nonstd::visit([](const auto& v) { return v.inputType; }, input);
+
+            switch (inputType) {
+            case InputBase::InputType::current_clamp: {
+                assert(nonstd::holds_alternative<SimulationConfig::InputLinear>(input) ||
```
WeinaJi, line 862: As I tested, `assert` here doesn't report error or abort the process. I don't know why. Perhaps better to throw with a clear message?

src/config.cpp
```diff
@@ -201,6 +202,172 @@ std::string toAbsolute(const fs::path& base, const fs::path& path) {
...
+
+        if (mean != valueIt.end()) {
+            parseOptional(valueIt, "mean", ret.mean);
+        }
+
+        if (mean_percent != valueIt.end()) {
+            parseOptional(valueIt, "mean_percent", ret.meanPercent);
+        }
+
+        if (ret.mean.has_value() && ret.meanPercent.has_value()) {
```
WeinaJi, line 310: Shall we combine the two exceptions? I think it is fine to throw as long as `ret.mean.has_value() == ret.meanPercent.has_value()`

src/config.cpp
```diff
@@ -201,6 +202,172 @@ std::string toAbsolute(const fs::path& base, const fs::path& path) {
...
+
+        parseOptional(valueIt, "variance", ret.variance);
+        return ret;
+    }
+    case Module::shot_noise: {
+        SimulationConfig::InputShotNoise ret;
+        parseCommon(ret);
+        parseMandatory(valueIt, "rise_time", debugStr, ret.riseTime);
+        parseMandatory(valueIt, "decay_time", debugStr, ret.decayTime);
+        parseOptional<int>(valueIt, "random_seed", ret.randomSeed, randomSeed);
```
WeinaJi, line 327: I remember that the template argument deduction should be working here (like in the `report` section), so no need to specify the types.

WeinaJi (commented):
The variants and the python binding looks good to me. Nice refactoring. Thanks!
Just an issue of the error reporting and two small comments that I have.

src/config.cpp
mgeplf: > I don't know why.

Was it a debug build? the `assert()` macro is a no-op in release.

This is more for debug builds, and only to make sure that when someone adds a new variant, it gets added in the right place.

src/config.cpp
mgeplf: I think this makes it more clear, so I'm going to keep it this way.

src/config.cpp
mgeplf: sure, fixed

src/config.cpp
WeinaJi: ah, I thought it was for detecting the wrong combination of a (valid) module type and a (valid) input type from users. Do we have a check for that ?

mgeplf: I don't think so, I've repurposed this test for that, and added a test.

WeinaJi (approved):
LGTM. Thanks for the changes.

@ferdonline , do you have more comments?

ferdonline (approved):

PR #203, opened 2022-05-27
Parse connection_overrides section

WeinaJi:


include/bbp/sonata/config.h
```diff
@@ -343,6 +343,32 @@ class SONATA_API SimulationConfig
...
+        /// node_set specifying postsynaptic nodes
+        std::string target;
+        /// Scalar to adjust synaptic strength, default = 1.
+        double weight{1.};
+        /// Rate to spontaneously trigger the synapses in this connection_override, default = None
+        nonstd::optional<double> spontMinis{nonstd::nullopt};
+        /// Snippet of hoc code to be executed on the synapses in this connection_override
+        std::string synapseConfigure;
+        /// Synapse helper files to instantiate the synapses in this connection_override
+        std::string modoverride;
```
NadirRoGue, line 362: LGTM, just a small remark,
Shouldn't these two fields be nonstd::optional as well? Just to be in sync with the other optionals that does not have a default value.
From what I see in the doc, modeoverride also indicates a prefix, which has a default value depending on synapse type, so optional would be safer since any use of this parameter when not setted will lead to an exception, whereas an empty string will lead to the wrong prefix.

WeinaJi: Good point. I've changed these two vars to `nonstd::optional` so that `default =  None` as in neurodamus.

mgeplf:
Thanks for "Add the function to list all connection_override names".

I think this is ready, @NadirRoGue, do you see anything I'm missing?

NadirRoGue:
Looks fine to me, thanks!

mgeplf (approved):

NadirRoGue (approved):

mgeplf:
Thanks @WeinaJi

PR #208, opened 2022-06-02
SimulationConfig parse target_simulator, node_sets_file and node_set

WeinaJi:


src/config.cpp
```diff
@@ -827,10 +832,41 @@ class SimulationConfig::Parser
...
+            return toAbsolute(_basePath, val);
+        } else {
+            try {
+                const auto circuitFile = parseNetwork();
+                const auto conf = CircuitConfig::fromFile(circuitFile);
+                return conf.getNodeSetsPath();
+            } catch (...) {
+                // Don't throw CircuitConfig exceptions in SimulationConfig and return empty string
+                return val;
+            }
```
WeinaJi, line 858: To be discussed with @mgeplf @ferdonline @jorblancoa 
1) The param `node_sets_file` is to override the one from circuit_config file, but it is optional in circuit_config with the default of empty string. What if no `node_sets_file` in circuit_config nor in simulation_config? Does SimulationConfig return empty string or `"node_sets.json"`? 
2) Here I set no throw, that SimulationConfig parser does not check the validity of the circuit config file.

mgeplf: Since we're using `optional` in several places, perhaps that's the way to go?

WeinaJi: Yes, I agree. Then the default value is also an empty string in SimulationConfig. Also in the case that `node_set` is not provided (meaning all non-virtual populations),  `node_sets_file` is not needed.

WeinaJi:
Hello, any more comment for this PR? Can we merge it ?

src/config.cpp
mgeplf: Probably worth getting other people's opinions, I think.

WeinaJi: @jorblancoa @ferdonline @NadirRoGue , what do you think about the two items above?

NadirRoGue: I prefer an empty string (or nullopt) to signal there is no file and to be consistent with CircuitConfig

WeinaJi: I have discussed with @ferdonline @jorblancoa offline, and they are both fine with the empty string as the default `node_sets_file`. Jorge raised the point that `reports:cells` should be optional as well, for which I opened a MR for discussion
https://bbpgitlab.epfl.ch/common/doc/circuit-documentation/-/merge_requests/24

For this PR, I think we can go ahead to merge.

mgeplf: > For this PR, I think we can go ahead to merge.

Ok, will do.

mgeplf (approved):

PR #209, opened 2022-06-13
for `biophysical` node populations, don't raise error if alternate_morphologies exists

mgeplf:
* fix python tests; unittest requires everything be declared in a class;
  can't have bare function
* hat-tip to @joni-herttuainen for noticing we weren't throwing an error
  when we should


joni-herttuainen (approved):
LGTM

PR #213, opened 2022-06-27
If population data was empty, default keys weren't being properly set

### LINKED ISSUE #205: should throw an error if biophysical_neuron_models_dir is missing for biophysical nodes
`it is needed only  for biophysical populations: "Path to the template HOC files defining the E-Model. There must be one defined for biophysical node populations.`

mgeplf:
* fixes #205


joni-herttuainen (approved):
Tested this on SNAP, seems to behave as expected now

PR #214, opened 2022-06-29
Teach SimulationConfig to return getExpandedJSON like CircuitConfig

mgeplf:
* remove the .getJSON method; one can just load the open themselves,
  path resolution is the hard part


NadirRoGue (approved):
LGTM

PR #217, opened 2022-07-07
Parse synapse mod variables in conditions/mechanisms section

WeinaJi:


WeinaJi (commented):
Hello @mgeplf @ferdonline @jorblancoa , any comment on this PR? 
And as we have agreed on the new format for synapse variables (BBPBGLIB-812), shall I remove the old format "synapses_init_depleted" from the parser?

python/generated/docstrings.h
```diff
@@ -875,19 +880,43 @@ static const char *__doc_bbp_sonata_SimulationConfig_Report_variableName = R"doc
...
+static const char *__doc_bbp_sonata_SimulationConfig_Run_IntegrationMethod_nicholson_ion = R"doc()doc";
+
+static const char *__doc_bbp_sonata_SimulationConfig_Run_SpikeLocation = R"doc()doc";
+
+static const char *__doc_bbp_sonata_SimulationConfig_Run_SpikeLocation_AIS = R"doc()doc";
+
+static const char *__doc_bbp_sonata_SimulationConfig_Run_SpikeLocation_invalid = R"doc()doc";
+
+static const char *__doc_bbp_sonata_SimulationConfig_Run_SpikeLocation_soma = R"doc()doc";
+
```
ferdonline, line 900: Do we need these? If yes, then they should have some description

python/generated/docstrings.h
```diff
@@ -875,19 +880,43 @@ static const char *__doc_bbp_sonata_SimulationConfig_Report_variableName = R"doc
...
 static const char *__doc_bbp_sonata_SimulationConfig_Run_dt = R"doc(Integration step duration in milliseconds)doc";
 
-static const char *__doc_bbp_sonata_SimulationConfig_Run_randomSeed = R"doc(Random seed)doc";
+static const char *__doc_bbp_sonata_SimulationConfig_Run_forwardSkip =
+R"doc(Run without Stimulus or Reports for given duration prior to t=0 using
+a timestep computed as dt=0.1*forward_skip. Default = None)doc";
```
ferdonline, line 905: Not sure I would bother too much about wrapping lines, it's also not great to lose indentation. Maybe if you wish
```suggestion
static const char *__doc_bbp_sonata_SimulationConfig_Run_forwardSkip =
    R"doc(Run without Stimulus or Reports for given duration prior to t=0 using a timestep computed as dt=0.1*forward_skip. Default = None)doc";
```

ferdonline (approved):
I think it looks great, just a couple of nitpicks which can totally be ignored

jorblancoa (approved):

python/generated/docstrings.h
mgeplf: These are machine generated, so don't need to be reviewed / changed.

mgeplf:
Looks good, thanks.

PR #224, opened 2022-09-15
Parse `metadata` and `beta_features` sections in SimulationConfig

WeinaJi:


include/bbp/sonata/config.h
```diff
@@ -618,6 +618,17 @@ class SONATA_API SimulationConfig
      */
     const nonstd::optional<std::string>& getNodeSet() const noexcept;
 
+    /**
+     * Returns the metadata section
+     */
+    const std::unordered_map<std::string, std::string>& getMetaData() const noexcept;
```
mgeplf, line 624: since `variantValueType` exists, maybe it could be used here too?

python/bindings.cpp
```diff
@@ -918,24 +918,56 @@ PYBIND11_MODULE(_libsonata, m) {
...
-        .def_property_readonly("target_simulator", &SimulationConfig::getTargetSimulator)
-        .def_property_readonly("node_sets_file", &SimulationConfig::getNodeSetsFile)
-        .def_property_readonly("node_set", &SimulationConfig::getNodeSet)
-        .def_property_readonly("list_report_names", &SimulationConfig::listReportNames)
-        .def_property_readonly("list_input_names", &SimulationConfig::listInputNames)
+        .def_static(
+            "from_file",
+            [](py::object path) { return SimulationConfig::fromFile(py::str(path)); },
+            DOC_SIMULATIONCONFIG(fromFile))
+        .def_property_readonly("base_path",
```
mgeplf, line 925: thanks for making this better.

include/bbp/sonata/config.h
WeinaJi: Unlike `beta_features` where various types of variables will be used for simulation, I intend that `metadata` holds only string variables as they are only for notes.  I've also addressed that in the sonata [spec](https://sonata-extension.readthedocs.io/en/latest/sonata_simulation.html#metadata).  
But I can change if it is better to allow different types of notes.

mgeplf: Ok, that works too - I'd prefer to keep the API relatively stable, so if you think it's going to remain that way, I'm fine with that.

The current parser will fail if a non-string is entered, so people should be aware of the problem.

mgeplf:
Ok, no other comments, let's get this in.

mgeplf (approved):

PR #226, opened 2022-09-19
don't make many small reads via HighFive/hdf5

mgeplf:
* use HighFive::ElementSet
* about twice as fast on my simple benchmark of reversing all ids: 
     sel = libsonata.Selection(list(reversed(sel_all.flatten()))) compared to reading everything as one large chunk

I don't know if hdf5 is coalescing reads - we may have to do that manually.

Note: This is still slow

### DIFF
--- a/src/population.hpp
+++ b/src/population.hpp
@@ -13,6 +13,11 @@
 
 #include <bbp/sonata/population.h>
 
+#include <algorithm>  // stable_sort, transform
+#include <iterator>   // back_inserter
+#include <numeric>    // iota
+#include <vector>
+
 #include <fmt/format.h>
 
 #include <highfive/H5File.hpp>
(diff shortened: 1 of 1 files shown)


matz-e (approved):

src/population.hpp
```diff
@@ -83,17 +89,35 @@ std::vector<T> _readSelection(const HighFive::DataSet& dset, const Selection& se
...
+    std::transform(ids_index.begin(),
+                   ids_index.end(),
+                   std::back_inserter(ids_sorted),
+                   [&ids](size_t i) { return static_cast<size_t>(ids[i]); });
+
+    std::vector<T> result(ids_sorted.size());
+    dset.select(HighFive::ElementSet{ids_sorted}).read(result.data());
+
+    std::vector<T> result1;
+    result1.resize(ids_sorted.size());
```
matz-e, line 115:
```suggestion
    std::vector<T> result1(ids_sorted.size());
```

mgeplf:
For [c5ce666](https://github.com/BlueBrain/libsonata/pull/226/commits/c5ce666e26dc94f9d36c2582e4306b55585a0271):
The speedup (from master) for all reversed ids is 24.3x, and for a random subset of ids (1/4 of all ids): the speedup is 17.5.

matz-e (approved):

PR #227, opened 2022-09-21
SimulationConfig parser

WeinaJi:
* remove forward_skip key from run (BBPBGLIB-913)
* remove minis_single_vesicle key from conditions (BBPBGLIB-915)
* add keys for additional seedings (BBPBGLIB-920)
   stimulus_seed, ionchannel_seed, minis_seed, synapse_seed
* add rs key in seclamp (BBPBGLIB-921)
* add keys neuromodulation_dtc and neuromodulation_strength in connection_overrides
(BBPBGLIB-885)

### DIFF
--- a/src/config.cpp
+++ b/src/config.cpp
@@ -390,6 +390,7 @@ SimulationConfig::Input parseInputModule(const nlohmann::json& valueIt,
         SimulationConfig::InputSeclamp ret;
         parseCommon(ret);
         parseMandatory(valueIt, "voltage", debugStr, ret.voltage);
+        parseOptional(valueIt, "rs", ret.rs, {0.01});
         return ret;
     }
     case Module::ornstein_uhlenbeck: {
@@ -836,7 +837,10 @@ class SimulationConfig::Parser
                       "integration_method",
                       result.integrationMethod,
                       {Run::IntegrationMethod::euler});
-        parseOptional(*runIt, "forward_skip", result.forwardSkip);
+        parseOptional(*runIt, "stimulus_seed", result.stimulusSeed, {0});
+        parseOptional(*runIt, "ionchannel_seed", result.ionchannelSeed, {0});
+        parseOptional(*runIt, "minis_seed", result.minisSeed, {0});
+        parseOptional(*runIt, "synapse_seed", result.synapseSeed, {0});
         return result;
     }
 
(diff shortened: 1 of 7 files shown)


src/config.cpp
```diff
@@ -914,7 +917,7 @@ class SimulationConfig::Parser
             parseOptional(valueIt, "enabled", report.enabled, {true});
 
             // Validate comma separated strings
-            const std::regex expr(R"(\w+(?:\s*,\s*\w+)*)");
+            const std::regex expr(R"(\w+.*(?:\s*,\s*\w+)*)");
```
mgeplf, line 920: do you mean `([\w.]+(?:\s*,\s*[\w.]+)*)`? the new regex covers things like:
`asdf,,,,,` which I don't think is intended.

We should probably add some more tests here, so it's clear what should pass, and what should fail.

WeinaJi: The old regex didn't support string like `AdEx.V_M` that's why I made the new regex `(\w+.*(?:\s*,\s*\w+)*)`. In the unit test, I've added `"variable_name": "AdEx.V_M,v"` that works fine. 
However, as you mentioned, `"variable_name": "asdf,,,,,,"` also passes the regex. @jorblancoa , is that fine to allow that? Can the reporting filter out those empty substring?

mgeplf: The idea is to reject incorrect strings at an early stage, so that we 'fail fast', and don't get far into running a sim without checking the config is correct.

Does the regex I suggested work?

WeinaJi: ah, ok, I see the problem in my change, it should be `[.]` that matches a dot.
The suggested ```([\w.]+(?:\s*,\s*[\w.]+)*)``` can reject `"AdEx.V_M,,,,"`, but it allows `".V_M"`. 
It seems ```(\w+[.]{0,1}\w+(?:\s*,\s*\w+[.]{0,1}\w+)*)``` can filter out  `".V_M"`, `"abcd,,,,"`. Is there a better expression?

mgeplf: I would prefer `(\w+\.?\w+(?:\s*,\s*\w+\.?\w+)*)`; and I think we should add what the expectations are. Is `AdEx.V_M.foo` allowed (for instance).

WeinaJi: `(\w+\.?\w+(?:\s*,\s*\w+\.?\w+)*)` is better, I will use that. Pramod confirms me that we could reject multiple dots such as`AdEx.V_M.foo`.

mgeplf: I noticed in the documentation (https://sonata-extension.readthedocs.io/en/latest/sonata_simulation.html#reports) that the last example is:
```
"variable_name": "i_membrane", "IClamp",
```
a) I don't think that it's valid json, so that should prob. be fixed in the docs
b) does `i_membrane` work at the moment?

WeinaJi: I just realized that `(\w+\.?\w+(?:\s*,\s*\w+\.?\w+)*)` doesn't allow single-letter variables, such as `"AdEx.V_M, v"`.
As I tested `(\w+(\.+\w+)?(?:\s*,\s*\w+(\.+\w+)?)*)` seems working.

mgeplf (approved):

Issue #218, opened 2022-08-08
Obtaining `"{nodes,edges}_file"` for a given population.


1uc:
We're using `libsonata` to parse a SONATA circuit config file. The information we'd like to get is the morphology directory, the `"nodes_file"` and `"edges_file"` for a given population.

I was hoping for an API similar to:
```
circuit_config = libsonata.CircuitConfig.from_file(config_file)
node_prop = config.node_population_properties(population="NodeA")
nodes_file = node_prop.nodes_file # or something similar
morphology_dir = node_prop.morphology_dir # This exists and is very convenient.
```
I was unable to find something like this in the documentation. Is this already possible? (Does this even make sense?)

Currently, we're getting the information my manually crawling the extended JSON.

1uc:
That's indeed the other place I looked for this functionality.

The downside of needing to open the file is that it must exist. While this is guaranteed in our immediate usecase (SpatialIndex). It might be nice to be able to use the circuit config file when creating the circuit, i.e. in TouchDetector or a little bit later in the pipeline, where the `edges_file` is created.

Second thought: if you add it to `Population` and you ever want to create a `Population` not by loading it from disk, e.g., because all the data already resides in memory, the value of `filename` would be something brittle, like `None`.

mgeplf:
> The downside of needing to open the file is that it must exist.

Yeah, I don't particularly like that idea.  Perhaps both `getting the paths` ways could exist. But for now, I can see the utility in having the `CircuitConfig` having the path available - I'll whip up a PR.

mgeplf:
@1uc are you happy w/ this? Can I close the issue?

PR #230, opened 2022-10-06
make h5 files available through PopulationProperties

mgeplf:
* circuit_config.json must now conform to the BBP standard for populations to be found
* the circuit_config.json now is the ground truth for available properties: if multiple nodes files contain the same population name, only the one referenced in the circuit_config.json is considered

### DIFF
--- a/include/bbp/sonata/config.h
+++ b/include/bbp/sonata/config.h
@@ -52,16 +52,20 @@ struct SONATA_API PopulationProperties {
 
     /**
      * Dictionary for alternate directory paths.
+     * It is discouraged to directly access the contents of the file.
+     * Instead use 'libsonata' to read this file.
      */
     std::unordered_map<std::string, std::string> alternateMorphologyFormats;
 
     /**
-     * Path to underlying elements H5 file.  It is discouraged to directly access this.
+     * Path to underlying elements H5 file.
      */
     std::string elementsPath;
 
     /**
-     * Path to underlying types csv file.  It is discouraged to directly access this.
+     * Path to underlying types csv file.
+     * It is discouraged to directly access the contents of the file.
+     * Instead use 'libsonata' to read this file.
      */
     std::string typesPath;
 };
--- a/include/bbp/sonata/config.h
+++ b/include/bbp/sonata/config.h
@@ -54,6 +54,16 @@ struct SONATA_API PopulationProperties {
      * Dictionary for alternate directory paths.
      */
     std::unordered_map<std::string, std::string> alternateMorphologyFormats;
+
+    /**
+     * Path to underlying elements H5 file.  It is discouraged to directly access this.
+     */
+    std::string elementsPath;
+
+    /**
+     * Path to underlying types csv file.  It is discouraged to directly access this.
+     */
+    std::string typesPath;
 };
 
 /**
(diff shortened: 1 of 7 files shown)


mgeplf:
fixes #218

joni-herttuainen (approved):
LGTM. Should we add a warning in snap about it being discouraged to access the file directly when getting the `.h5_filepath`?

mgeplf:
> LGTM. Should we add a warning in snap about it being discouraged to access the file directly when getting the `.h5_filepath`?

We can decide that when we update the snap code - for now, I think just putting it in the docstring is fine

src/config.cpp
```diff
@@ -134,6 +134,47 @@ namespace {
 // to be replaced by std::filesystem once C++17 is used
 namespace fs = ghc::filesystem;
 
+void checkBiophysicalPopulations(
+    const std::unordered_map<std::string, PopulationProperties>& populations) {
+    for (const auto& it : populations) {
+        const auto population = it.first;
```
1uc, line 140: Will this take a copy; and if yes, is that intentional?

python/generated/docstrings.h
```diff
@@ -230,10 +230,18 @@ static const char *__doc_bbp_sonata_PopulationProperties_alternateMorphologyForm
 
 static const char *__doc_bbp_sonata_PopulationProperties_biophysicalNeuronModelsDir = R"doc(Path to the template HOC files defining the E-Mode)doc";
 
+static const char *__doc_bbp_sonata_PopulationProperties_elementsPath =
+R"doc(Path to underlying elements H5 file. It is discouraged to directly
+access this.)doc";
```
1uc, line 235: Maybe consider:

```
It is discouraged to directly access the contents of the file. Instead use 'libsonata' to read this file.
```

Personally, I would draw the line at manually opening/reading the file, rather than even knowing the path of the file that contains a particular population.

1uc (approved):
Many thanks, I've built this version on BB5 and used it to get the nodes file for one of the SONATA usecases. It works perfectly. I'm looking forward to removing substantial complexity from SpatialIndex.

I've carefully checked the C++ it looks fine. There's two very minor comments. The features are covered by tests, with the minor exception that `typesPath` strictly isn't check; I don't see any reason why it would be incorrect.

src/config.cpp
mgeplf: tnx, I'll fix that

python/generated/docstrings.h
mgeplf: yup, your wording is more clear, I will fix that.

(for future ref the `docstrings.h` is generated, so I hope you weren't reviewing it too carefully :))

include/bbp/sonata/config.h
```diff
@@ -52,16 +52,20 @@ struct SONATA_API PopulationProperties {
...
     /**
      * Dictionary for alternate directory paths.
+     * It is discouraged to directly access the contents of the file.
+     * Instead use 'libsonata' to read this file.
      */
     std::unordered_map<std::string, std::string> alternateMorphologyFormats;
 
     /**
-     * Path to underlying elements H5 file.  It is discouraged to directly access this.
+     * Path to underlying elements H5 file.
```
1uc, line 61: Does this one get the same disclaimer?

1uc (approved):

include/bbp/sonata/config.h
mgeplf: grr, bad copy pasta, fixed.

1uc (approved):

PR #232, opened 2022-10-10
Fix #226 for empty selections.

matz-e:
Part of our software uses empty selections to extract the datatype from the returned Numpy arrays. But with #226, HDF5 does not like the created empty HDF5 slabs, so return directly when nothing is requested.


GianlucaFicarelli:
We could add also a test with empty selections in `test_source_nodes` and `test_target_nodes` as in https://github.com/BlueBrain/libsonata/pull/233 (I opened that PR because I didn't see this one)

WeinaJi (approved):

WeinaJi:
It seems that the new tag also breaks the synapse reading in our simulation CIs. 
https://bbpgitlab.epfl.ch/hpc/sim/blueconfigs/-/jobs/384346
We need this fix asap.

PR #238, opened 2022-10-25
refactor python tests so they are not in a monolithic file

mgeplf:


matz-e (approved):
Test count seems to not have changed.

PR #241, opened 2022-10-31
Allow for partial circuit configs

mgeplf:
* a partial circuit config has a `metadata` property `type` that is equal to `partial`
* this relaxes some checking that is done on circuit configs:
  * can be missing the top level `networks` key
  * can be missing the `nodes` and `edges` properties under `network`
* mandatory properties aren't enforced (for example biophysical properties don't need a `[alternative_]morphologies_dir`

### DIFF
--- a/include/bbp/sonata/config.h
+++ b/include/bbp/sonata/config.h
@@ -76,7 +76,20 @@ struct SONATA_API PopulationProperties {
 class SONATA_API CircuitConfig
 {
   public:
-    enum class Type { invalid, Complete, Partial };
+    enum class ConfigStatus {
+        /// needed for parsing json contents that are null / not an enum value
+        invalid,
+        /// all mandatory properties exist, and the the config should return
+        /// correct values in all possible cases
+        Complete,
+        /**
+         * Partial configs relax the requirements:
+             - can be missing the top level networks key
+             - can be missing the nodes and edges properties under network
+             - mandatory properties aren't enforced (for example for biophysical circuits)
+         */
+        Partial
+    };
 
     /**
      * Parses a SONATA JSON config file.
@@ -103,9 +116,10 @@ class SONATA_API CircuitConfig
     static CircuitConfig fromFile(const std::string& path);
 
     /**
-     * Returns the path to the node sets file.
+     * Returns the `completeness` of the checks that are performed for the
+     * circuit; see `ConfigStatus` for more information
      */
-    Type getCircuitConfigType() const;
+    ConfigStatus getCircuitConfigStatus() const;
 
     /**
      * Returns the path to the node sets file.
@@ -181,8 +195,8 @@ class SONATA_API CircuitConfig
     // manifest variables
     std::string _expandedJSON;
 
-    // Config Type
-    Type _circuitType = Type::Complete;
+    /// How strict we are checking the circuit config
+    ConfigStatus _status = ConfigStatus::Complete;
 
     // Path to the nodesets file
     std::string _nodeSetsFile;
--- a/include/bbp/sonata/config.h
+++ b/include/bbp/sonata/config.h
@@ -76,6 +76,8 @@ struct SONATA_API PopulationProperties {
 class SONATA_API CircuitConfig
 {
   public:
+    enum class Type { invalid, Complete, Partial };
+
     /**
      * Parses a SONATA JSON config file.
      *
@@ -99,6 +101,11 @@ class SONATA_API CircuitConfig
      *          - Multiple populations with the same name in different edge/node networks
      */
     static CircuitConfig fromFile(const std::string& path);
+
+    /**
+     * Returns the path to the node sets file.
+     */
+    Type getCircuitConfigType() const;
 
     /**
      * Returns the path to the node sets file.
@@ -174,11 +181,14 @@ class SONATA_API CircuitConfig
     // manifest variables
     std::string _expandedJSON;
 
+    // Config Type
+    Type _circuitType = Type::Complete;
+
     // Path to the nodesets file
     std::string _nodeSetsFile;
 
-    // Default components value for all networks
-    Components _components;
+    // Nodes network paths
+    std::vector<SubnetworkFiles> _networkNodes;
 
     // Node populations that override default components variables
     std::unordered_map<std::string, PopulationProperties> _nodePopulationProperties;
--- a/include/bbp/sonata/config.h
+++ b/include/bbp/sonata/config.h
@@ -81,14 +81,14 @@ class SONATA_API CircuitConfig
         invalid,
         /// all mandatory properties exist, and the the config should return
         /// correct values in all possible cases
-        Complete,
+        complete,
         /**
          * Partial configs relax the requirements:
              - can be missing the top level networks key
              - can be missing the nodes and edges properties under network
              - mandatory properties aren't enforced (for example for biophysical circuits)
          */
-        Partial
+        partial
     };
 
     /**
@@ -196,7 +196,7 @@ class SONATA_API CircuitConfig
     std::string _expandedJSON;
 
     /// How strict we are checking the circuit config
-    ConfigStatus _status = ConfigStatus::Complete;
+    ConfigStatus _status = ConfigStatus::complete;
 
     // Path to the nodesets file
     std::string _nodeSetsFile;
--- a/python/tests/test_config.py
+++ b/python/tests/test_config.py
@@ -31,7 +31,7 @@ class TestCircuitConfig(unittest.TestCase):
 
         self.assertEqual(self.config.edge_population('edges-AB').name, 'edges-AB')
 
-        self.assertEqual(self.config.config_status, CircuitConfigStatus.Complete)
+        self.assertEqual(self.config.config_status, CircuitConfigStatus.complete)
 
     def test_expanded_json(self):
         config = json.loads(self.config.expanded_json)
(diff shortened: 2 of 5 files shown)


joni-herttuainen (dismissed):

python/tests/test_config.py
```diff
@@ -57,6 +61,47 @@ def test_get_population_properties(self):
...
+            "metadata": { "type": "partial" },
+            "networks-mispelled": { },
+            }
+        cc = CircuitConfig(json.dumps(contents), PATH)
+        self.assertEqual(cc.type, CircuitConfigType.Partial)
+        self.assertEqual(cc.node_populations, set())
+        self.assertEqual(cc.edge_populations, set())
+
+        contents = {
+            "metadata": { "type": "partial" },
```
eleftherioszisis, line 84: partial is a property that is defined on a per population basis because we can have a mix of complete circuits with partial populations in various building steps, especially when the ngv is considered.

We can also have it at the root level so that libsonata treats it with extra care, however we also need the per population partial definition so that we know which population is complete and which isn't.

For example:
```json
{
  "version": 2,
  "manifest": {
    "$BASE_DIR": "."
  },
  "networks": {
    "nodes": [
      {
        "nodes_file": "...",
        "populations": {
          "root_neurons": {
            "type": "biophysical",
            "partial": [
              "cell-properties",
              "morphologies",
              "emodels"
            ],
            "morphologies_dir": "...",
            "biophysical_neuron_models_dir": "..."
          }
        }
      }
    ]
  }

```

src/config.cpp
```diff
@@ -716,46 +740,57 @@ CircuitConfig::CircuitConfig(const std::string& contents, const std::string& bas
...
                 }
             }
         };
+
     updateDefaultProperties(_nodePopulationProperties, "biophysical");
     updateDefaultProperties(_edgePopulationProperties, "chemical_synapse");
 
-    checkBiophysicalPopulations(_nodePopulationProperties);
+    if (_circuitType == Type::Complete) {
+        checkBiophysicalPopulations(_nodePopulationProperties);
```
1uc, line 782: I can't spot any side effects in `checkBiophysicalPopulations`. Therefore, doing this conditionally should be correct.

src/config.cpp
```diff
@@ -716,46 +740,57 @@ CircuitConfig::CircuitConfig(const std::string& contents, const std::string& bas
...
+
+                if (component.alternateMorphologyFormats.empty()) {
+                    component.alternateMorphologyFormats = components.alternateMorphologiesDir;
                 }
-                if (entry.second.biophysicalNeuronModelsDir.empty()) {
-                    entry.second.biophysicalNeuronModelsDir =
-                        _components.biophysicalNeuronModelsDir;
+
+                if (component.biophysicalNeuronModelsDir.empty()) {
+                    component.biophysicalNeuronModelsDir = components.biophysicalNeuronModelsDir;
```
1uc, line 769: I can't spot the difference between the `lhs` and `rhs`. Is it possible the `rhs` should be `_components.bio...`?

src/config.cpp
```diff
@@ -716,46 +740,57 @@ CircuitConfig::CircuitConfig(const std::string& contents, const std::string& bas
...
-                    entry.second.type = defaultType;
+                auto& component = entry.second;
+                if (component.type.empty()) {
+                    component.type = defaultType;
                 }
-                if (entry.second.alternateMorphologyFormats.empty()) {
-                    entry.second.alternateMorphologyFormats = _components.alternateMorphologiesDir;
+
+                if (component.alternateMorphologyFormats.empty()) {
+                    component.alternateMorphologyFormats = components.alternateMorphologiesDir;
```
1uc, line 765: I can't spot the difference between the `lhs` and `rhs`. Is it possible the `rhs` should be `_components.bio...`?

src/config.cpp
```diff
@@ -716,46 +740,57 @@ CircuitConfig::CircuitConfig(const std::string& contents, const std::string& bas
...
-                        _components.biophysicalNeuronModelsDir;
+
+                if (component.biophysicalNeuronModelsDir.empty()) {
+                    component.biophysicalNeuronModelsDir = components.biophysicalNeuronModelsDir;
                 }
-                if (entry.second.morphologiesDir.empty()) {
-                    entry.second.morphologiesDir = _components.morphologiesDir;
+
+                if (component.morphologiesDir.empty()) {
+                    component.morphologiesDir = components.morphologiesDir;
```
1uc, line 773: Potentially once more the comment about `lhs == rhs`.

src/config.cpp
```diff
@@ -53,6 +53,11 @@ struct adl_serializer<nonstd::optional<T>> {
 namespace bbp {
 namespace sonata {
 
+NLOHMANN_JSON_SERIALIZE_ENUM(CircuitConfig::Type,
+                             {{CircuitConfig::Type::invalid, nullptr},
+                              {CircuitConfig::Type::Partial, "partial"},
+                              {CircuitConfig::Type::Complete, "complete"}})
```
1uc, line 59: It might be worth checking if the capitalization is consistent. We have `Type::invalid` lower case `i`, then `Type::Partial` but `SpikeLocation::soma`. The string identifiers are all lower case.

include/bbp/sonata/config.h
```diff
@@ -76,6 +76,8 @@ struct SONATA_API PopulationProperties {
 class SONATA_API CircuitConfig
 {
   public:
+    enum class Type { invalid, Complete, Partial };
```
1uc, line 79: You could reuse some of the MR documentation to explain what complete and partial mean:

> * a partial circuit config has a `metadata` property `type` that is equal to `partial`
> * this relaxes some checking that is done on circuit configs:
>   * can be missing the top level `networks` key
>   * can be missing the `nodes` and `edges` properties under `network`
> * mandatory properties aren't enforced (for example biophysical
>   properties don't need a `[alternative_]morphologies_dir`

include/bbp/sonata/config.h
```diff
@@ -174,11 +181,14 @@ class SONATA_API CircuitConfig
     // manifest variables
     std::string _expandedJSON;
 
+    // Config Type
+    Type _circuitType = Type::Complete;
```
1uc, line 185: I feel a better name than `Type` should exist. I'll update the comment if I find something.

1uc (commented):
I only have very superficial comments. Please ignore them at will.

src/config.cpp
mgeplf: yeah, perhaps it should be name `checkInvariantsBiophysicalPopulations` or `throwIfNonconformingPopulations`?

src/config.cpp
mgeplf: `component` vs `components` (plural); I agree it's too subtle - I will change it.

src/config.cpp
mgeplf: yeah, I didn't like the difference in case either, but I was following what had been done before.

I need to check if `NLOHMANN_JSON_SERIALIZE_ENUM` is case sensitive; sadly the spec has mixed case, so we may be stuck :(

include/bbp/sonata/config.h
mgeplf: agreed.

include/bbp/sonata/config.h
mgeplf: I also dislike it; I had `CircuitConfigType`, but that seemed redundant.  `Filled` / `Completeness`? I'm also not coming up with anything satisfactory.

src/config.cpp
1uc: Not on my account. I'm only saying that `checkBiophysicalPopulations` might have done something like also set a default for example, but I looked and that's not the case, as far as I can tell.

include/bbp/sonata/config.h
1uc: OTOH, it looks like what the setting really affects, is if the file is checked to be conforming/complete. So maybe the setting could be related to checking: `CheckCompleteness`, `Validate`, `QuickFail`, `AssertMode`, `LazyChecking`, `AllowMissing`.

include/bbp/sonata/config.h
matz-e: Why is `invalid` lowercase, when all others are capitalized? This somehow really bugs me.

include/bbp/sonata/config.h
matz-e: `Status`? Seems not really a `type` qualifier, but really something about the _state_ of the circuit.

include/bbp/sonata/config.h
mgeplf: > Why is invalid lowercase, when all others are capitalized?

See https://github.com/BlueBrain/libsonata/pull/241#discussion_r1010176908

>  This somehow really bugs me.

Same.

src/config.cpp
mgeplf: I can confirm that `NLOHMANN_JSON_SERIALIZE_ENUM` is case sensitive...

include/bbp/sonata/config.h
matz-e: Ah, missed the spec part :(

python/tests/test_config.py
mgeplf: That's a good point; however, I think for the initial version, we can make the whole thing partial, and then work towards making individual parts partial.

eleftherioszisis: Sure, that's fine with me!

include/bbp/sonata/config.h
mgeplf: > You could reuse some of the MR documentation to explain what complete and partial mean:

Agreed, will add doc.

include/bbp/sonata/config.h
mgeplf: I went with `ConfigStatus`; not the happiest with that, but hopefully it conveys the meaning.

matz-e (approved):

matz-e (approved):

mgeplf:
Will merge once https://github.com/BlueBrain/sonata-extension/pull/14 has been approved (in case there are changes)

jdcourcol:
@jplanasc You asked for that some time ago. (Sometimes we want to visualize unfinished circuits)

PR #247, opened 2023-02-10
Enable setting the C++ standard.

1uc:
Currently, `libsonata` contains a hard coded `CMAKE_CXX_STANDARD=14`. This prevents users from compiling libsonata as a C++ project. `libsonata` uses `optional-lite` which provides fallbacks for `std::optional` called `nonstd::optional`. The actual type of `nonstd::optional` differs depending on the language features supported by the compiler, e.g. if C++17 is available, `nonstd::optional` is an alias for `std::optional`. This causes linking issues when compiling `libsonata` as a C++14 library and using it in a C++17 application.

One way of resolving this issue is to allow `libsonata` to be built as a C++17 project. This commit makes the hard coded `CMAKE_CXX_STANDARD` a variable, but sets it to `14` (the lowest supported version of the standard) unless requested otherwise.


1uc:
Currently, `-DCMAKE_CXX_STANDARD=20` doesn't work due to deprecation warnings which are treated as errors. However, I suggest that be tackled in a separate PR.

matz-e (approved):

mgeplf (approved):

PR #250, opened 2023-02-13
Spatial index

mgeplf:


include/bbp/sonata/config.h
```diff
@@ -183,12 +200,6 @@ class SONATA_API CircuitConfig
         std::string biophysicalNeuronModelsDir;
     };
 
-    struct SubnetworkFiles {
```
edasubert, line 186: why is this removed?

edasubert (approved):
seems ok to me, but it is long time since I worked with c++

include/bbp/sonata/config.h
mgeplf: good eye; it wasn't used, and should have been removed before.

mgeplf:
> seems ok to me, but it is long time since I worked with c++

Same here :)

I just wanted to make sure someone else looked at the names, to make sure it was what we decided.

matz-e (approved):

PR #255, opened 2023-03-27
Multi clause nodeset

mgeplf:


src/node_sets.cpp
```diff
@@ -325,118 +353,137 @@ int64_t get_integer_or_throw(const json& el) {
...
 
-NodeSetRules _dispatch_node(const json& contents) {
-    NodeSetRules ret;
+NodeSetRulePtr _dispatch_node(const std::string& attribute, const json& value) {
+    if (value.is_number()) {
+        if (attribute == "population") {
+            throw SonataError("'population' must be a string");
+        }
+
+        int64_t v = get_integer_or_throw(value);
```
NadirRoGue, line 362: I guess a function to parse unsigned integers can be added in the future? Not that it will overflow but just a comment.

NadirRoGue (dismissed):
LGTM

src/node_sets.cpp
mgeplf: Nope, good call, I will fix that.

NadirRoGue (approved):
LGTM

Issue #260, opened 2023-04-17
CircuitConfig.from_file may fail with generic RuntimeError and cause OOM


GianlucaFicarelli:
When loading a circuit from a wrong big (4.5GB) file by mistake, as in this example:

```python
import libsonata

input_path = "/gpfs/bbp.cscs.ch/project/proj134/workflow-outputs/genrich-test/morphologyAssignmentConfig/root/nodes.h5"
cc = libsonata.CircuitConfig.from_file(input_path)
```
The result is:
```
Traceback (most recent call last):
  File "./tmp/try_libsonata.py", line 4, in <module>
    cc = libsonata.CircuitConfig.from_file(input_path)
RuntimeError
12.17user 5.02system 0:17.29elapsed 99%CPU (0avgtext+0avgdata 9285064maxresident)k
0inputs+0outputs (0major+18023minor)pagefaults 0swaps
```

Instead, in other cases of invalid json, you can get something like:
```
RuntimeError: [json.exception.parse_error.101] parse error at line 1, column 1: syntax error while parsing value - invalid literal; last read: 'm'
```

Possible issues:
- the first RuntimeError doesn't say that the file is invalid
- the file is fully read in memory and it may cause an OOM

1uc:
> the file is fully read in memory and it may cause an OOM
When I looked at it, it was most likely this one.

mgeplf:
I have a fix in progress, I shouldn't be reading the full contents of the file :)

mgeplf:
This is fixed by #261

PR #261, opened 2023-04-17
Don't read the full node sets file before parsing

mgeplf:


mgeplf:
fixes #260

src/node_sets.cpp
```diff
@@ -73,6 +77,21 @@ class NodeSets
         parse_compound(j, node_sets_);
     }
 
+    explicit NodeSets(const fs::path& path)
+        : NodeSets(json::parse(std::fstream(path))) {}
```
1uc, line 81: Unlike `fromFile` this doesn't check that the file exists. Maybe something
```
const fs::path& validate_path(fs::path& path)
     if (!fs::exists(path)) {
        throw SonataError(fmt::format("Path does not exist: {}", std::string(path)));
    }
    return path;
}

explicit NodeSets(const fs::path& path)
    : NodeSets(json::parse(std::fstream(validate_path(path)))) {}


static std::unique_ptr<NodeSets> fromFile(const std::string& path_) {
    return std::make_unique<detail::NodeSets>(path);
}
```

Finally, probably pointless comment but: `json::parse` only needs an `std::ifstream`.

src/node_sets.cpp
```diff
@@ -73,6 +77,21 @@ class NodeSets
...
+
+    explicit NodeSets(const std::string& content)
+        : NodeSets(json::parse(content)) {}
```
1uc, line 84: This will allow `detail::NodeSets("nodesets.h5")` to compile, it won't parse (because of missing inner `"`).

1uc (approved):
Only minor comments, please feel free to ignore them.

matz-e (approved):

src/node_sets.cpp
mgeplf: I'm not sure I follow this point, could you expand?  Is there a test I should add that would catch what you're suggesting?

1uc: I tend to not distinguish between path object and strings. Often a string is a perfectly fine way of specifying a path. Hence when I read the first ctor, I think: "Ah, it accepts paths." Good, let's pass it a `"nodeset.h5"`, resulting in something like:
```
auto detail_nodeset = detail::NodeSet("foo.h5");
```
thinking that this will be the same as `detail::NodeSet(fs::path("foo.h5"))`. It's not. The slightly unfortunate thing is that it'll compile. The good news is that it'll fail, because `"foo.h5"` is not a valid JSON object `"\"foo.h5\""` would be.

I understand it's all happening inside `detail` so it's not user facing. Hence there's likely little cause to have a path which is an `std::string`.

Is this a bit more clear? (There's nothing deep or serious happening here.)

mgeplf: Ah, thank you, yes, that's perfectly clear now.

I also see strings as paths (bad habits die hard), but since it's internal, I'm going to leave it as it is.

The best plan would have been to not have any ctors, and just a `fromFile`, `fromContents`, to avoid this ambiguity.

matz-e (approved):

PR #265, opened 2023-05-03
nlohmann is_array() for empty arrays has type `null` instead of array

mgeplf:


WeinaJi:
Thanks for making the fix. I also notice it from the neurodamus CI tests. Currently I remove those empty "connection_overrides" arrays in the sonata config files to avoid the false negative. Will this fix go to the next tag? I assume that I can still deploy `0.1.19` in bb5 ?

PR #268, opened 2023-05-17
Include Catch2 only if building tests

alkino:


olupton (approved):
LGTM

Some background: there were failures when we tried to change the Spack recipe for `libsonata` to not depend on `catch2` when the `tests` variant is not set. It's important to be able to do this, because other projects want to move to `catch2@3` and `libsonata` is not compatible with that.

mgeplf:
Cool, thanks for the PR and the context.

PR #271, opened 2023-05-31
Add 'lfp' report type and 'electrodes_file' field

jorblancoa:
I also updated the documentation here:
https://github.com/BlueBrain/sonata-extension/pull/11

### DIFF
--- a/src/config.cpp
+++ b/src/config.cpp
@@ -977,6 +977,11 @@ class SimulationConfig::Parser
         parseOptional(*runIt, "minis_seed", result.minisSeed, {0});
         parseOptional(*runIt, "synapse_seed", result.synapseSeed, {0});
         parseOptional(*runIt, "electrodes_file", result.electrodesFile, {""});
+
+        if (!result.electrodesFile.empty()) {
+            result.electrodesFile = toAbsolute(_basePath, result.electrodesFile);
+        }
+
         return result;
     }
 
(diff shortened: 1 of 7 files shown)


src/config.cpp
```diff
@@ -868,6 +869,7 @@ class SimulationConfig::Parser
         parseOptional(*runIt, "ionchannel_seed", result.ionchannelSeed, {0});
         parseOptional(*runIt, "minis_seed", result.minisSeed, {0});
         parseOptional(*runIt, "synapse_seed", result.synapseSeed, {0});
+        parseOptional(*runIt, "electrodes_file", result.electrodesFile, {""});
```
WeinaJi, line 872: I think it is safer to parse it to be an absolute path here, like the other file paths in the config file.

jorblancoa: The optional files are parsed in the same way. For example:

`parseOptional(*outputIt, "log_file", result.logFile, {""});
 parseOptional(*outputIt, "spikes_file", result.spikesFile, {"out.h5"});`

WeinaJi (approved):
LGTM. Let's wait to merge the SPEC first https://github.com/BlueBrain/sonata-extension/pull/11

jorblancoa:
Now that the spec has been merged, maybe we can also merge this? @mgeplf

PR #275, opened 2023-06-06
implement spec for node population properties for vasculature, astroc…

mgeplf:
…yte, endfeet

ie:
    vasculatureFile
    vasculatureMesh
    endfeetMeshesFile
    microdomainsFile
    spineMorphologiesDir

### DIFF
--- a/python/bindings.cpp
+++ b/python/bindings.cpp
@@ -543,13 +543,28 @@ PYBIND11_MODULE(_libsonata, m) {
         m, "NodePopulationProperties", "Stores node population-specific network information")
         .def_readonly("spatial_segment_index_dir",
                       &NodePopulationProperties::spatialSegmentIndexDir,
-                      DOC_NODE_POPULATION_PROPERTIES(spatialSegmentIndexDir));
+                      DOC_NODE_POPULATION_PROPERTIES(spatialSegmentIndexDir))
+        .def_readonly("vasculature_file",
+                      &NodePopulationProperties::vasculatureFile,
+                      DOC_NODE_POPULATION_PROPERTIES(vasculatureFile))
+        .def_readonly("vasculature_mesh",
+                      &NodePopulationProperties::vasculatureMesh,
+                      DOC_NODE_POPULATION_PROPERTIES(vasculatureMesh))
+        .def_readonly("microdomains_file",
+                      &NodePopulationProperties::microdomainsFile,
+                      DOC_NODE_POPULATION_PROPERTIES(microdomainsFile))
+        .def_readonly("spine_morphologies_dir",
+                      &NodePopulationProperties::spineMorphologiesDir,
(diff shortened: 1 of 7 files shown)


python/bindings.cpp
```diff
@@ -543,13 +543,28 @@ PYBIND11_MODULE(_libsonata, m) {
         m, "NodePopulationProperties", "Stores node population-specific network information")
         .def_readonly("spatial_segment_index_dir",
                       &NodePopulationProperties::spatialSegmentIndexDir,
-                      DOC_NODE_POPULATION_PROPERTIES(spatialSegmentIndexDir));
+                      DOC_NODE_POPULATION_PROPERTIES(spatialSegmentIndexDir))
+        .def_readonly("vasculature_file",
+                      &NodePopulationProperties::vasculatureFile,
+                      DOC_NODE_POPULATION_PROPERTIES(vasculatureFile))
+        .def_readonly("vasculature_mesh",
```
eleftherioszisis, line 550: This should be "vasculature_mesh_file"

mgeplf: yup, I'll change that; can you just use this name for the test?

eleftherioszisis: My bad. The test config is old. vasculature_mesh is the correct term, being also in the spec.

python/tests/test_config.py
```diff
@@ -151,6 +151,33 @@ def test_biophysical_properties_raises(self):
...
+                "vasculature_mesh": "$BASE_DIR/vasculature_mesh.h5",
+            },
+            "networks": {
+                "nodes": [
+                    {
+                        "nodes_file": "$BASE_DIR/nodes1.h5",
+                        "populations": {
+                            "nodes-A": {
+                                "type": "vasculature",
+                                "vasculature_file": "$BASE_DIR/vasculature_file.h5",
```
eleftherioszisis, line 169: The `vasculature_mesh` is below `vasculature_file` in ngv configs. The components are not usually used.

src/config.cpp
```diff
@@ -655,29 +694,69 @@ class CircuitConfig::Parser
         result.biophysicalNeuronModelsDir = getJSONPath(components,
                                                         "biophysical_neuron_models_dir");
 
+        result.vasculatureFile = getOptionalJSONPath(components, "vasculature_file");
+        result.vasculatureMesh = getOptionalJSONPath(components, "vasculature_mesh");
+        result.endfeetMeshesFile = getOptionalJSONPath(components, "endfeetMeshesFile");
```
eleftherioszisis, line 699: Why is this in camel case, while the rest in snake case?

python/tests/test_config.py
mgeplf: yeah, I'm checking the resolution works, if it is in the config.

src/config.cpp
mgeplf: tnx, fixed

eleftherioszisis (approved):

PR #276, opened 2023-06-16
Prevent node_set materialization for `@libary` with integers

mgeplf:
* If a node_set contains something like:
 {"layer1": {"layer": 1} ...  # < note it's 1 and not a string "1"

this would previously resolved if "layer" was a set of @library strings,
because it would have matched the numeric 1 to the enum-like values stored
in the /nodes/.../0/layer and not /nodes/.../0/@library/layer


mgeplf:
@WeinaJi can you have a look at this? I'd like to get it merged before we do a release.

WeinaJi (approved):
LGTM, thanks!

PR #277, opened 2023-06-22
Extend metadata type to be int, float, bool and string

WeinaJi:
Fixes https://github.com/BlueBrain/sonata-extension/issues/26

### DIFF
--- a/tests/test_config.cpp
+++ b/tests/test_config.cpp
@@ -584,7 +584,7 @@ TEST_CASE("SimulationConfig") {
         CHECK(overrides[1].neuromodulationStrength == 0.75);
         REQUIRE_THAT(config.getMetaData(),
                      Catch::Matchers::Predicate<decltype(config.getMetaData())>(
-                         [](auto& metadata) -> bool {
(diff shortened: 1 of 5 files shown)


mgeplf (approved):

mgeplf:
thanks!

tests/test_config.cpp
```diff
@@ -583,7 +583,7 @@ TEST_CASE("SimulationConfig") {
         CHECK(overrides[1].neuromodulationDtc == 100);
         CHECK(overrides[1].neuromodulationStrength == 0.75);
 
-        CHECK(config.getMetaData().size() == 2);
+        CHECK(config.getMetaData().size() == 4);
```
mgeplf, line 586: can you add a test here to get the values?

mgeplf (changes requested):

tests/test_config.cpp
WeinaJi: Yes, I've just push the change.

mgeplf (dismissed):

mgeplf (approved):

Issue #279, opened 2023-06-27
DeprecationWarning with numpy 1.25: Conversion of an array with ndim > 0 to a scalar is deprecated


GianlucaFicarelli:
With numpy 1.25 and libsonata 0.1.21, this code causes a DeprecationWarning:

```python
import numpy as np
import libsonata
cc = libsonata.CircuitConfig.from_file("tests/data/config/circuit_config.json")
pop = cc.edge_population("edges-AB")
ids = np.asarray([1])
pop.afferent_edges(ids)
```

Result:
```python
<ipython-input-1-d0433371129e>:6: DeprecationWarning: Conversion of an array with ndim > 0 to a scalar is deprecated, and will error in future. Ensure you extract a single element from your array before performing this operation. (Deprecated NumPy 1.25.)
  pop.afferent_edges(ids)
Selection([(0, 1), (2, 4)])
```

While these calls don't cause any warning:
```python
In [2]: ids = [1]
   ...: pop.afferent_edges(ids)
   ...: 
Selection([(0, 1), (2, 4)])

In [3]: ids = 1
   ...: pop.afferent_edges(ids)
Selection([(0, 1), (2, 4)])
```

mgeplf:
Thanks for the repro, I can generate the same error.

I will look into it.

mgeplf:
It's an ordering issue when creating the bindings; because the scalar version is registered first, it is used.  I guess previously, auto-coercion would happen from the single element array, to a scalar (since this came first), where we would change it back into a vector...

I need to check where else we may be doing this.

PR #280, opened 2023-06-30
Move std::vector methods above scalar methods in bindings

mgeplf:
* Otherwise, numpy coercion kicks in, and an argument like `np.asarray([42])` gets coerced to 42, the scalar binding is used, and then we convert it to a std::vector internally.
* not only is this inefficient, newer numpy versions emit:

    DeprecationWarning: Conversion of an array with ndim > 0 to a scalar is deprecated, and will error in future.
    Ensure you extract a single element from your array before performing this operation.
    (Deprecated NumPy 1.25.)


mgeplf:
fixes #279

matz-e (approved):

PR #281, opened 2023-06-30
test for completely empty datasets

mgeplf:
* prevents failures with cryptic error message:
    HDF5-DIAG: Error detected in HDF5 (1.10.4) thread 140737340061504:
    #000: ../../../src/H5Shyper.c line 7057 in H5Sselect_hyperslab(): hyperslab doesn't support H5S_NULL space

### DIFF
--- a/src/population.hpp
+++ b/src/population.hpp
@@ -57,14 +57,6 @@ std::set<std::string> _listExplicitEnumerations(const HighFive::Group h5Group,
     return names;
 }
 
-bool _isEmtpyOrNullDataset(const HighFive::DataSet& dset) {
-    //  Once HighFive releases a version with
-    // https://github.com/BlueBrain/HighFive/pull/787
-    // can switch to the more clear:
-    // return dset.getElementCount() == 0;
-    return dset.getStorageSize() == 0;
-}
-
 template <typename T>
 std::vector<T> _readChunk(const HighFive::DataSet& dset, const Selection::Range& range) {
     std::vector<T> result;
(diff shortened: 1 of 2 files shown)


src/population.hpp
```diff
@@ -29,15 +29,15 @@ namespace sonata {
 
 namespace {
 
-const char* H5_DYNAMICS_PARAMS = "dynamics_params";
-const char* H5_LIBRARY = "@library";
+const char* const H5_DYNAMICS_PARAMS = "dynamics_params";
+const char* const H5_LIBRARY = "@library";
```
1uc, line 33: On the pedantic level of things: This looks like it might be a `constexpr [const] char [const] * const`.

src/population.hpp
```diff
@@ -89,7 +92,7 @@ std::vector<T> _readSelection(const HighFive::DataSet& dset, const Selection& se
 
 template <typename T, typename std::enable_if<std::is_pod<T>::value>::type* = nullptr>
 std::vector<T> _readSelection(const HighFive::DataSet& dset, const Selection& selection) {
-    if (selection.ranges().empty()) {
+    if (selection.ranges().empty() || dset.getElementCount() == 0) {
```
1uc, line 95: We'll need to keep track of the `dset.getElementCount() == 0` (which is a collective property) vs. `selection.ranges().empty()` (which is not) during the upcoming refactoring.

1uc (approved):

src/population.hpp
mgeplf: ok

1uc (approved):

PR #282, opened 2023-07-07
disable py37

mgeplf:


setup.py
adrien-berchet, line 1: We should also change numpy version to `>=1.17.3`, the first to have a wheel for Py38

mgeplf: thanks, good point - fixed

eleftherioszisis (approved):

PR #284, opened 2023-07-17
Add missing reversal property of shotnoise family

jorblancoa:
Implement change in spec:
https://github.com/BlueBrain/sonata-extension/pull/30


src/config.cpp
```diff
@@ -444,6 +444,7 @@ SimulationConfig::Input parseInputModule(const nlohmann::json& valueIt,
         parseMandatory(valueIt, "rise_time", debugStr, ret.riseTime);
         parseMandatory(valueIt, "decay_time", debugStr, ret.decayTime);
         parseOptional(valueIt, "random_seed", ret.randomSeed);
+        parseOptional(valueIt, "reversal", ret.reversal);
```
mgeplf, line 447: can we make it clear what the default value is here, so it's not relying on the ctor?

mgeplf: yes please, if the default is missing other places, fix that too :)

mgeplf (approved):

PR #285, opened 2023-07-26
Return a structured array for the spikes in python

jorblancoa:
- Modify Spike object from std::pair to struct
 
 Following the discussions in [[BBPBGLIB-1044]](https://github.com/BlueBrain/neurodamus/pull/8#top)

### DIFF
--- a/include/bbp/sonata/report_reader.h
+++ b/include/bbp/sonata/report_reader.h
@@ -27,6 +27,7 @@ struct SONATA_API DataFrame {
 
 using Spike = std::pair<NodeID, double>;
 using Spikes = std::vector<Spike>;
+using SpikesArrays = std::pair<std::vector<NodeID>, std::vector<double>>;
 
 /// Used to read spike files
 class SONATA_API SpikeReader
@@ -53,6 +54,10 @@ class SONATA_API SpikeReader
                    const nonstd::optional<double>& tstart = nonstd::nullopt,
                    const nonstd::optional<double>& tstop = nonstd::nullopt) const;
 
+        SpikesArrays get_arrays(const nonstd::optional<Selection>& node_ids = nonstd::nullopt,
+                                const nonstd::optional<double>& tstart = nonstd::nullopt,
+                                const nonstd::optional<double>& tstop = nonstd::nullopt) const;
+
         /**
          * Return the way data are sorted ('none', 'by_id', 'by_time')
          */
@@ -62,6 +67,8 @@ class SONATA_API SpikeReader
         Population(const std::string& filename, const std::string& populationName);
 
         Spikes spikes_;
+        std::vector<NodeID> node_ids_;
+        std::vector<double> timestamps_;
         Sorting sorting_ = Sorting::none;
         // Use for clamping of user values
         double tstart_, tstop_;
--- a/include/bbp/sonata/report_reader.h
+++ b/include/bbp/sonata/report_reader.h
@@ -25,7 +25,15 @@ struct SONATA_API DataFrame {
     std::vector<float> data;
 };
 
-using Spike = std::pair<NodeID, double>;
+typedef struct {
+    NodeID node_id;
+    double timestamp;
+} Spike;
+
+bool operator==(const Spike& lhs, const Spike& rhs) {
+    return lhs.node_id == rhs.node_id && lhs.timestamp == rhs.timestamp;
+}
+
 using Spikes = std::vector<Spike>;
 
 /// Used to read spike files
--- a/include/bbp/sonata/report_reader.h
+++ b/include/bbp/sonata/report_reader.h
@@ -48,15 +48,23 @@ class SONATA_API SpikeReader
         std::tuple<double, double> getTimes() const;
 
         /**
+         * Create the spikes from the vectors of node_ids and timestamps
+         */
+        Spikes createSpikes() const;
+
+        /**
          * Return spikes with all those node_ids between 'tstart' and 'tstop'
          */
         Spikes get(const nonstd::optional<Selection>& node_ids = nonstd::nullopt,
                    const nonstd::optional<double>& tstart = nonstd::nullopt,
                    const nonstd::optional<double>& tstop = nonstd::nullopt) const;
 
-        SpikesArrays get_arrays(const nonstd::optional<Selection>& node_ids = nonstd::nullopt,
-                                const nonstd::optional<double>& tstart = nonstd::nullopt,
-                                const nonstd::optional<double>& tstop = nonstd::nullopt) const;
+        /**
+         * Return the node_ids and timestamps vectors with all node_ids between 'tstart' and 'tstop'
+         */
+        SpikesArrays getArrays(const nonstd::optional<Selection>& node_ids = nonstd::nullopt,
+                               const nonstd::optional<double>& tstart = nonstd::nullopt,
+                               const nonstd::optional<double>& tstop = nonstd::nullopt) const;
 
         /**
          * Return the way data are sorted ('none', 'by_id', 'by_time')
@@ -66,7 +74,6 @@ class SONATA_API SpikeReader
       private:
         Population(const std::string& filename, const std::string& populationName);
 
-        Spikes spikes_;
         std::vector<NodeID> node_ids_;
         std::vector<double> timestamps_;
         Sorting sorting_ = Sorting::none;
--- a/include/bbp/sonata/report_reader.h
+++ b/include/bbp/sonata/report_reader.h
@@ -25,15 +25,15 @@ struct SONATA_API DataFrame {
     std::vector<float> data;
 };
 
-typedef struct {
+struct SpikeStruct {
     NodeID node_id;
     double timestamp;
-} Spike;
-
-bool operator==(const Spike& lhs, const Spike& rhs) {
-    return lhs.node_id == rhs.node_id && lhs.timestamp == rhs.timestamp;
-}
-
+    bool operator==(const SpikeStruct& other) const {
+        return node_id == other.node_id && timestamp == other.timestamp;
+    }
+};
+using SpikeStruct = struct SpikeStruct;
+using Spike = std::pair<NodeID, double>;
 using Spikes = std::vector<Spike>;
 
 /// Used to read spike files
--- a/include/bbp/sonata/report_reader.h
+++ b/include/bbp/sonata/report_reader.h
@@ -28,9 +28,6 @@ struct SONATA_API DataFrame {
 struct SpikeStruct {
     NodeID node_id;
     double timestamp;
-    bool operator==(const SpikeStruct& other) const {
-        return node_id == other.node_id && timestamp == other.timestamp;
-    }
 };
 using SpikeStruct = struct SpikeStruct;
 using Spike = std::pair<NodeID, double>;
--- a/include/bbp/sonata/report_reader.h
+++ b/include/bbp/sonata/report_reader.h
@@ -27,7 +27,10 @@ struct SONATA_API DataFrame {
 
 using Spike = std::pair<NodeID, double>;
 using Spikes = std::vector<Spike>;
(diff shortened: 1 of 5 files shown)


mgeplf:
> I'm however not very happy with the convoluted path of the data, which I know it's not from this PR, but I would expect better from an "optimized library". 

I expect there needs to be a new c++ method that builds these separated vectors which is waht I was getting at here: https://github.com/BlueBrain/neurodamus/pull/8/files#discussion_r1273085740

> Besides, in `get()`, there is an unconditional copy of all spikes for a filtering that is hardly requested. [report_reader.cpp#L136](https://github.com/BlueBrain/libsonata/blob/c05694dc3333299541aba27e2ebcdbfc6bc6ded5/src/report_reader.cpp#L136C27-L136C27)

That copy should be fixed.  The filtering, however, is important.

> Maybe we should go back to requirements. Do we need the get() interface as is? is it really worth to have columnar data for data which is invariably two columns?

On the C++ side, I think it's fine; keeping that API stable is important.  On the python side, a list of tuples is not great, but API stability is more important.

Once we have the new python methods in place, we can start issuing `warnings` that people may get better performance from switching to it.  Then, perhaps, we can deprecate it in the future.

python/bindings.cpp
```diff
@@ -1173,13 +1173,69 @@ PYBIND11_MODULE(_libsonata, m) {
...
+
     py::class_<SpikeReader::Population>(m, "SpikePopulation", "A population inside a SpikeReader")
         .def("get",
              &SpikeReader::Population::get,
              DOC_SPIKEREADER_POP(get),
              "node_ids"_a = nonstd::nullopt,
              "tstart"_a = nonstd::nullopt,
              "tstop"_a = nonstd::nullopt)
+        .def(
+            "get_array",
```
mgeplf, line 1186: I don't think we need both the `get_array` and the `get_dict`, since they are nearly equivalent.  Let's just go for the `get_dict` one.

jorblancoa:
Shall we merge this in order to create a tag together with the LFP changes?

mgeplf:
> I'm however not very happy with the convoluted path of the data, which I know it's not from this PR, but I would expect better from an "optimized library".

I think @ferdonline's point of " I'm however not very happy with the convoluted path of the data, which I know it's not from this PR, but I would expect better from an "optimized library"." still stands.  I think we should do it with a less convoluted data path before it gets committed.

jorblancoa:
Does that belong to this PR?

mgeplf:
I would say so; the extra array created by https://github.com/BlueBrain/libsonata/pull/285/files#diff-cc07100b7c7235ddf263fda0d515a7b40ed1fe5b871714e0ab5c1ec5298ddb0dR1193 can be avoided if some internal methods were added to get the id and timestamp separately

jorblancoa:
> I would say so; the extra array created by https://github.com/BlueBrain/libsonata/pull/285/files#diff-cc07100b7c7235ddf263fda0d515a7b40ed1fe5b871714e0ab5c1ec5298ddb0dR1193 can be avoided if some internal methods were added to get the id and timestamp separately

Makes sense, but then the best would be to have them as members in the report_reader.h right? Otherwise the spikes_ need to be used anyway to filter and then converted to the 2 arrays.

mgeplf:
> Makes sense, but then the best would be to have them as members in the report_reader.h right?

Yeah, I think I understand what you mean and that would make sense.

jorblancoa:
Since we cant reuse the existing code for filtering and fernando's usecase only needs the 2 raw arrays, would you be ok for that? @ferdonline @mgeplf

include/bbp/sonata/report_reader.h
```diff
@@ -62,6 +67,8 @@ class SONATA_API SpikeReader
         Population(const std::string& filename, const std::string& populationName);
 
         Spikes spikes_;
+        std::vector<NodeID> node_ids_;
+        std::vector<double> timestamps_;
```
ferdonline, line 71: As we are now keeping the raw arrays in memory I think we can drop the `spikes_` struct and build it on the fly when sorting/filtering is needed

ferdonline (approved):

mgeplf:
Don't we still have the `convoluted path of the data` which was of the main things that needed to be improved?
It looks like `getArrays` calls `get`, which makes an unconditional copy of all spikes by `createSpikes` creating pairs from the underlying node_id and timestamps.
Back in these pairs are destructured into two new arrays.  Can’t we create arrays of only the data that is required (ie: based on the gids/times), so they aren’t large, and create them in the format that is required?  Having an Iterator would probably make that easier.

ferdonline:
@mgeplf IIUC at [line 162](https://github.com/BlueBrain/libsonata/pull/285/files#diff-1f8cee3dc53f35c903106e4c6ba100226dca70687e1a9cd79d1a5850ba5bf25eR162) there's the fast path. If the client wants to filter data then the copy happens to `vector<pair>` which IMO is fair since that layout is ideal for sorting together. We could potentially further optimize the case of only filtering without ordering and avoid the copy, but I don't think we have that use case now, so I'd agree to leave as is and touch at a later stage.

mgeplf:
But we want to be able to filter things; that’s an important use case. It seems weird to me to have users of a library have to know that there is a `fast-path` if the client wants to sort the data rather than doing the right thing on the library side.

On the neurodamus side, isn’t it useful to only load spikes that are going to be used in the simulation rather than loading all of them?  For instance, if the simulation is going to run for 1s, but the spikes file contains 10s worth of spikes?  What about the case where the simulation is only doing a subset of node_ids?  Wouldn’t filtering the data before it all gets loaded be valuable?  These are the sorts of optimizations that can be done here, and then the would benefit everyone using the library.

They’re also useful right now, as when people do analysis, a subset of the data is usually looked at.

jorblancoa:
> Don't we still have the `convoluted path of the data` which was of the main things that needed to be improved?
> It looks like `getArrays` calls `get`, which makes an unconditional copy of all spikes by `createSpikes` creating pairs from the underlying node_id and timestamps.
> Back in these pairs are destructured into two new arrays. Can’t we create arrays of only the data that is required (ie: based on the gids/times), so they aren’t large, and create them in the format that is required? Having an Iterator would probably make that easier.

It makes sense, I wanted to reuse code but is true in case of calling getArrays it doesnt make sense to call get() to make the pairs only for the filtering. I pushed some changes to filter directly in the getArrays() method.

ferdonline:
> On the neurodamus side, isn’t it useful to only load spikes that are going to be used in the simulation rather

In neurodamus we can't always filter ahead of time because of setup like coreneuron+save-restore, so it was ok. But I get it that maybe for other uses we want filters more often.

python/tests/test_reports.py
```diff
@@ -48,6 +48,13 @@ def test_get_spikes_from_population(self):
 
         self.assertEqual(len(self.test_obj['All'].get(node_ids=[])), 0)
 
+        dict_data = self.test_obj['All'].get_dict()
+        self.assertTrue((dict_data["node_ids"] == np.asarray([5, 2, 3, 2, 3])).all())
+        self.assertTrue((dict_data["timestamps"] == np.asarray([0.1, 0.2, 0.3, 0.7, 1.3])).all())
+        dict_data_filtered = self.test_obj['All'].get_dict(node_ids=[5, 3], tstart=0.2)
+        self.assertTrue((dict_data_filtered["node_ids"] == np.asarray([3, 3])).all())
+        self.assertTrue((dict_data_filtered["timestamps"] == np.asarray([0.3, 1.3])).all())
+
```
mgeplf, line 57: could you add tests for just using tstart/tstop?

I happend to run this:
```
import libsonata
sr=libsonata.SpikeReader('spikes.h5')
sr['All'].get_dict(tstart=0)
```

And got:
```
RuntimeError: bad optional access
```

mgeplf:
Comparing the `get_dict` to the `get()` method, it seems like there is a performance regression when getting a subset of ids:
```
    def read_all_subset_nodes():
        sp = sr['All']
        node_ids = libsonata.Selection(ids)
        spikes = sp.get(node_ids)

    def read_all_subset_nodes_structure():
        sp = sr['All']
        node_ids = libsonata.Selection(ids)
        spikes = sp.get_dict(node_ids)

    print("read_all_subset_nodes")
    timeit(read_all_subset_nodes)

    print("read_all_subset_nodes_structure")
    timeit(read_all_subset_nodes_structure)
```

Gives:

```
************* spikes-decimated-unsorted.h5
read_all_subset_nodes
1 loop, best of 5: 130.22718537412584 per loop
read_all_subset_nodes_structure
1 loop, best of 5: 972.7277841931209 per loop

************* spikes-decimated-sorted-by-time.h5
read_all_subset_nodes
1 loop, best of 5: 130.0559630645439 per loop
read_all_subset_nodes_structure
1 loop, best of 5: 972.6187489759177 per loop

************* spikes-decimated-sorted-by-ids.h5
read_all_subset_nodes
1 loop, best of 5: 96.04782155901194 per loop
read_all_subset_nodes_structure
1 loop, best of 5: 972.36246633064 per loop
```

I'm not a big fan of different access mechanisms having much different execution profiles, because that means library consumers have to benchmark everything to know what to use - which isn't very ergonomic.

In this case, we'll have to fix it later.

include/bbp/sonata/report_reader.h
```diff
@@ -46,13 +47,25 @@ class SONATA_API SpikeReader
          */
         std::tuple<double, double> getTimes() const;
 
+        /**
+         * Create the spikes from the vectors of node_ids and timestamps
+         */
+        Spikes createSpikes() const;
```
mgeplf, line 53: I think this should be a private function; since we don't want to have to support more API than necesary.

include/bbp/sonata/report_reader.h
```diff
@@ -27,6 +27,7 @@ struct SONATA_API DataFrame {
 
 using Spike = std::pair<NodeID, double>;
 using Spikes = std::vector<Spike>;
+using SpikesArrays = std::pair<std::vector<NodeID>, std::vector<double>>;
```
mgeplf, line 30: I'd use a `struct` here to make it more future proof; if we decide to add additional info, the calling code can be recompiled without changes.

perhaps:
```
struct SpikeTimes{
  std::vector<NodeID> nodeIds;
  std::vector<double> timestamps;
};
```

include/bbp/sonata/report_reader.h
```diff
@@ -61,7 +74,8 @@ class SONATA_API SpikeReader
       private:
         Population(const std::string& filename, const std::string& populationName);
 
-        Spikes spikes_;
+        std::vector<NodeID> node_ids_;
```
mgeplf, line 77: This could become a `SpikeTimes`, instead of two arrays; then the `getArrays` could be have two calling options: one without any arguments returns a `const &` to `SpikeTimes`, so no copies are required of the C++ side.

mgeplf:
> Comparing the get_dict to the get() method, it seems like there is a performance regression when getting a subset of ids:

I should also add that the fast path (ie: `get_dict()`) is much faster: (new vs old): 0.36195594631135464 vs 9.994007629342377, so it's very worthwhile.

mgeplf:
Nice, fixing the regression makes a big difference:
```
************* spikes-decimated-unsorted.h5
read_all_subset_nodes
1 loop, best of 5: 131.17122020898387 per loop
read_all_subset_nodes_structure
1 loop, best of 5: 129.58019144897116 per loop
************* spikes-decimated-sorted-by-time.h5
read_all_subset_nodes
1 loop, best of 5: 131.06627149000997 per loop
read_all_subset_nodes_structure
1 loop, best of 5: 129.91921105398796 per loop
************* spikes-decimated-sorted-by-ids.h5
read_all_subset_nodes
1 loop, best of 5: 96.16886089701438 per loop
read_all_subset_nodes_structure
1 loop, best of 5: 129.54882664396428 per loop
```

mgeplf (approved):

mgeplf:
Thanks @jorblancoa!

PR #289, opened 2023-09-21
Fix toJSON incorrect empty, node_id and compound problems

mgeplf:


mgeplf:
@joni-herttuainen can you try this version? the wheels should be available to you here: https://github.com/BlueBrain/libsonata/actions/runs/6259270581?pr=289

joni-herttuainen (approved):
Yep, this seems to have fixed the issue

PR #294, opened 2023-10-18
Use `RawIndex = std::vector<std::array<.., 2>>`.

1uc:
The memory layout of an `std::vector<std::array<...>>` is better than `std::vector<std::vector<...>>`. This allows HighFive to avoid numerous allocations and copies. Additionally, I feel it's simpler because it avoid the ambiguity of rows of differing length, when in fact the `RawIndex` seems to be a vector of ranges.

### DIFF
--- a/src/edge_index.cpp
+++ b/src/edge_index.cpp
@@ -11,6 +11,7 @@
 
 #include <bbp/sonata/common.h>
 
+#include <array>
 #include <cstdint>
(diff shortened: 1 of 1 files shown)


matz-e (approved):

mgeplf (approved):

mgeplf:
very nice, thanks

PR #295, opened 2023-10-18
Formatting CI prints diff on failure.

1uc:
As a not very active contributor, it would be convenient if CI told me where I forgot to add that one `" "`. This can be faster than finding out which version of `clang-format` libsonata uses, installing it and then formatting the code. Naturally, in slightly more complex cases one simply needs to install `clang-format`. This mainly decreases the hurdle to submit tiny fixups of typos and the like.

If it doesn't interfere with other peoples workflow, I suggest printing the git diff.

### DIFF
--- a/ci/check_clang_format.sh
+++ b/ci/check_clang_format.sh
@@ -23,5 +23,8 @@ changes=$(git-clang-format 'HEAD~1' $DIRS_TO_FORMAT)
 if [[ $(echo "$changes" | grep -n1 'changed files') ]]; then
     echo "The following files require changes to pass the current clang-format"
     echo "$changes"
+    echo ""
+    echo "This is the diff:"
+    git diff
     exit 1
 fi


1uc:
Currently, the PR contains a commit which intentionally breaks formatting.

matz-e (approved):

src/edge_index.cpp
```diff
@@ -13,8 +13,8 @@
 
 #include <cstdint>
 #include <set>
-#include <unordered_map>
 #include <vector>
+#include <unordered_map>
```
matz-e, line 17:
```suggestion
#include <unordered_map>
#include <vector>
```

matz-e (dismissed):
I like it, but it needs fixes for the CI to pass.

mgeplf:
good call

mgeplf (approved):

PR #296, opened 2023-10-18
Add `CMAKE_BUILD_PARALLEL_LEVEL=2` when building wheels.

1uc:


1uc:
Seems to have zero effect on the build time.

mgeplf:
> Seems to have zero effect on the build time.

Huh, I'm also surprised.  Thanks for trying.

PR #298, opened 2023-10-23
Optimize loading `?fferent_edges`.

1uc:
Optimized reading of edge IDs by aggregating ranges into larger (GPFS-friendly) ranges before creating the appropriate HDF5 selection to reduce the number of individual reads. Then it filters out any unneeded data in memory. This is very similar to work done in #183.

This PR introduces the following:
* Extend the internal API for merging `libsonata.Selection`.
* Add internal API for reading in bulk and filtering.
* Load `?fferent_edge` in bulk.
* Compile-time constant `SONATA_PAGESIZE` which controls how large the merged region need to be.

### DIFF
--- a/src/edge_index.cpp
+++ b/src/edge_index.cpp
@@ -17,6 +17,7 @@
 #include <unordered_map>
 #include <vector>
 
+#include "read_bulk.hpp"
 
 namespace bbp {
 namespace sonata {
@@ -52,7 +53,6 @@ const HighFive::Group targetIndex(const HighFive::Group& h5Root) {
     }
     return h5Root.getGroup(TARGET_INDEX_GROUP);
 }
-
 
 Selection resolve(const HighFive::Group& indexGroup, const NodeID nodeID) {
     if (nodeID >= indexGroup.getDataSet(NODE_ID_TO_RANGES_DSET).getSpace().getDimensions()[0]) {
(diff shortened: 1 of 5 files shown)


1uc:
This is a WIP. The intention is to create an short-term release that can be used in simulations that have a large impact on GPFS. Therefore, the following topics are out-of-scope:

* Memory usage in pathological cases.
* Code organization, e.g. moving functionality to neurodamus.
* Docstrings and unit-testing.

src/edge_index.cpp
```diff
@@ -15,6 +15,11 @@
 #include <set>
 #include <unordered_map>
 #include <vector>
+#include <array>
+
+#include <nlohmann/json.hpp>
+#include <fstream>
+#include <unistd.h>
```
matz-e, line 22: `unistd.h` inclusion should be guarded for Windows :)

1uc: used to generate a filename for some debug output.

1uc: Let me stress once more that while this does reduce the time to read synapses by a factor 8 - 10. This isn't mergeable.

1uc: No more `<unistd.h>` :)

1uc:
When reading edge ids for a large file we get a 10'000x speedup. The benchmark is:
```
n_nodes = 100
n_gids = 10000000
n_ranks = n_nodes * 40
gid_stride = 2 # simulates some selection effect.

all_edge_ids = []
for k_rank in range(n_ranks):
    gids = np.arange(*fair_chunk(n_ranks, k_rank, n_gids), gid_stride)
    edge_ids = population.afferent_edges(gids)
    all_edge_ids.append(edge_ids.ranges)
```
This would compute the edge IDs used by each MPI rank for analysis purposes. With the optimization it takes about 2-4s without the first 10 rank take 77s (or 8.5h total).

1uc:
The PR uses templates to hide the difference between `Selection::Ranges` which is an `std::vector<std::pair>` and the `RawIndex` which an `std::vector<std::array>`. We need to use `std::get<.>` to access `x.first` and `x[0]` in a uniform manner.

1uc:
The solution only works for `afferent_edges` but not for `efferent_edges`, due to locality.

sergiorg-hpc (dismissed):

matz-e (approved):

src/population.hpp
```diff
@@ -22,6 +22,8 @@
 
 #include <highfive/H5File.hpp>
 
+#include "read_bulk.hpp"
```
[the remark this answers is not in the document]
1uc: @matz-e @joni-herttuainen fixed.

joni-herttuainen (approved):

PR #303, opened 2023-11-06
Simplify single group check.

1uc:


src/population.hpp
```diff
@@ -143,11 +143,7 @@ struct Population::Impl {
               h5Root.getGroup("0").exist(H5_DYNAMICS_PARAMS)
                   ? _listChildren(h5Root.getGroup("0").getGroup(H5_DYNAMICS_PARAMS))
                   : std::set<std::string>{}) {
-        size_t groupID = 0;
-        while (h5Root.exist(std::to_string(groupID))) {
-            ++groupID;
-        }
-        if (groupID != 1) {
+        if (h5Root.exist("1")) {
```
matz-e, line 146: This seems like a funny check, because it assumes that groups are sequential. I don't see any such requirement in the SONATA spec. Seems like we should construct a set of all subgroups and ensure its only element is `0`.

1uc: It's the same check as before. Improving the check is a different task. This simply makes it obvious to the reader that the check is "flawed", and therefore they can decide if it's worth their time to fix it.

joni-herttuainen: Agreed. The check may need to be changed but that is a separate issue.

matz-e (approved):

PR #306, opened 2023-11-07
Move `Selection` to its own header.

1uc:
The new header is `<bbp/sonata/selection.h>` and it's included in `<bbp/sonata/population.h`. This ensures doesn't affect which include provide `Selection` in an adverse way.


1uc:
Splitting the two help avoid cyclic problems, e.g. if one want to add a class that requires `Selection` but is itself used in `Population`. Then without splitting the header the new class must be at the right position in `population.h`.

PR #307, opened 2023-11-07
Inject dataset reading via `Hdf5Reader`.

1uc:
The idea is that one injects call-backs for reading selections from datasets. These callbacks implement:

```
template<class T>
std::vector<T> readSelection(const HighFive::DataSet& dset, Selection& selection);
HighFive::File open_file(const std::string& filename);
```

The default calls `_readSelection`. Separate readers can implement variants suitable for their purpose, e.g. MPI-IO. The advantage is that the reader has control over both the required collective semantics; and setting HDF5 properties. This allows some readers to have non-collective behaviour, such as short-circuiting for empty selections. While others implement strictly collective reading of datasets.

Additionally, we can create a suite of readers for different usecases, e.g. MPI-IO for neurodamus, aggregating reader for serial workloads with GPFS (and a separate one for LUSTER if need be), the default implementation because it minimizes number of bytes read.


python/bindings.cpp
```diff
@@ -403,6 +412,30 @@ void bindReportReader(py::module& m, const std::string& prefix) {
...
+    if (import_mpi4py() < 0) {
+        throw std::runtime_error("Could not load mpi4py API.");
+    }
+#endif
+
+    py::class_<Hdf5Reader>(m, "Hdf5Reader").def(py::init([]() { return Hdf5Reader(); }));
+
+    m.def("make_collective_reader",
+          [](py::object comm, bool collective_metadata, bool collective_transfer) {
+#if SONATA_HAS_MPI == 1
```
ferdonline, line 425: When `SONATA_HAS_MPI` is False I would rather support we dont even expose collective stuff, as it might be misleading to ask for a collective reader and get a normal one.

ferdonline:
Interesting design. As I understand you are implementing a collective reader already, so in Neurodamus we would need basically to call `make_collective_reader` and use it as argument in calls to the config which return storage. I think this is fine.

python/bindings.cpp
1uc: Summary: I think it's important that MPI is entirely optional in libsonata, even when writing code that wants to use collective MPI-IO. At most, I'd add a flag that prevents the returning a suitable default.

The lines aren't precise because likely we'll move anything MPI to its own repo. The reason is that `mpi4py` is a build dependency; and one can't have optional build dependencies (nicely).

Nevertheless, I think code that's written for collective IO should work even if the user fails to install `mpi4py`. HDF5 doesn't fail either if it can't do collective IO, instead it returns the correct values and sets an internal flag users can check. Functionally collective IO is optional. It's only important for performance reasons under certain conditions.

The risk is that installing `mpi4py` or `libsonata` with MPI support will fail to build under certain conditions. I suspect many of our users will be unable or unwilling to debug these failures. To me it's a big difference if what I'm running is a large production run that requires MPI-IO to run reasonably fast, or if it's something small (debugging) that runs fast simply because it's small. Large runs are always done on a cluster and we can expect someone adept at these issues to install the required toolchain (including MPI-enabled libsonata), even if not, there's a necessity and therefore it makes sense to spend time to debug why something failed to build. Small runs are carried out anywhere, including laptops where MPI might not be installed or it might not work correctly at that moment, etc. In my opinion I don't want to force users to have to figure these issues out just to run something that takes `2.3s` with MPI-IO and `2.2s` without MPI-IO. Similarly, I don't want every downstream project to remember to implement a graceful fallback.

To detect if the important applications are using MPI-IO we can:
* check GPFS waiters,
* run Darshan,
* use `ROMIO_PRINT_HINTS=1` and check the output, e.g., during module testing on the BB5.

Therefore, I think, on BB5, we're in a sufficiently strong position to notice if it's silently not using collective IO when it really should be.

include/bbp/sonata/hdf5_reader.h
```diff
@@ -0,0 +1,66 @@
...
+  public:
+    virtual std::vector<T> readSelection(const HighFive::DataSet& dset,
+                                         const Selection& selection) const = 0;
+};
+
+template <class Tuple>
+class Hdf5PluginInterface;
+
+template <class... Types>
+class Hdf5PluginInterface<std::tuple<Types...>>: virtual public Hdf5PluginReadInterface<Types>...
```
matz-e, line 24: While I'm not particularly fond of this way of doing things, it's a lot cleaner IMO than using macros to instantiate all the readers. Cheers!

matz-e (commented):
Nice! If I get this correct, we can just pull out the collective read interface into it's own library?

include/bbp/sonata/hdf5_reader.h
1uc: That's my favourite part! It's a virtual template method =)

1uc:
> Nice! If I get this correct, we can just pull out the collective read interface into it's own library?

Yes, almost certainly, because mpi4py is a build dependency; and one can't have optional build dependencies (nicely).

include/bbp/sonata/hdf5_reader.h
matz-e: My issue is more with C++ itself. But this uses less C, so I'm happy :smile:

1uc:
This PR has been heavily reworked since the previous round of discussions. Eventually it will only include the `Hdf5Reader` API. The two commits that refactor libsonata have been split into their own PRs. The implementation of an MPI-IO reader has been moved to a separate PR (until it can be moved to its own repository).

include/bbp/sonata/hdf5_reader.h
```diff
@@ -0,0 +1,195 @@
...
+                                                               const AltRanges& xsel,
+                                                               const Selection& ysel) const = 0;
+
+    virtual std::vector<std::array<uint64_t, 2>> readSelection(const HighFive::DataSet& dset,
+                                                               const Selection& xsel,
+                                                               const AltRanges& ysel) const = 0;
+
+    virtual std::vector<std::array<uint64_t, 2>> readSelection(const HighFive::DataSet& dset,
+                                                               const AltRanges& xsel,
+                                                               const AltRanges& ysel) const = 0;
```
1uc, line 55: After #319 this reduces down to one method.

src/read_bulk.hpp
```diff
@@ -43,6 +43,18 @@ size_t flatSize(const std::vector<Range>& ranges) {
...
+    std::vector<Int> result;
+    result.reserve(flatSize(ranges));
+    for (const auto& range : ranges) {
+        for (auto v = std::get<0>(range); v < std::get<1>(range); ++v) {
+            result.emplace_back(v);
+        }
+    }
+    return result;
+}
+
```
1uc, line 57: This kinda of annoyance goes away with #319.

1uc:
The questions for review are:
1. Are the rules regarding collectiveness of libsonata acceptable?
2. Do we need the ability to pass options to the reader on `getAttribute` and similar? Currently, I can only think of `collective` which would be used to specify if the call is collective. Things like block size, I fell would go into the reader.
3. Naming of `Hdf5Reader`.

The next review would include question like:
1. Can we have an optional dependency `libsonata[mpi]`? The advantage is that `libsonata` can control the version of `libsonata_mpi` it's compatible with, which is useful if the API of `Hdf5Reader` needs to change.
2. Can we have wrappers for `libsonata_mpi.make_collective_reader` in `libsonata`? The advantage is that one could then always do `libsonata.make_collective_reader` which would always succeed. If `libsonata_mpi` is available it will return a collective reader, otherwise it'll return the default reader.

With the above, code written to use collective IO, should automatically run even if the user can't install `mpi4py` or `libsonata_mpi`. Therefore, users can avoid the pain of installing the difficult dependencies; and devs don't need to worry about making it easy for their users.

include/bbp/sonata/hdf5_reader.h
```diff
@@ -0,0 +1,197 @@
...
+      virtual public Hdf5PluginRead2DInterface<Us>...
+{
+  public:
+    /// Open the HDF5 file.
+    ///
+    /// This allows setting File Access Properties.
+    virtual HighFive::File openFile(const std::string& path) const = 0;
+};
+
+/// Abstraction for reading HDF5 datasets.
```
mgeplf, line 77: thanks for the clear docs and examples of right and wrong

include/bbp/sonata/hdf5_reader.h
```diff
@@ -0,0 +1,197 @@
...
+
+    /// Create a valid Hdf5Reader with the default plugin.
+    Hdf5Reader();
+
+    /// Create an Hdf5Reader with a user supplied plugin.
+    Hdf5Reader(std::shared_ptr<Hdf5PluginInterface<supported_1D_types, supported_2D_types>> impl);
+
+    /// Read the selected subset of the one-dimensional array.
+    ///
+    /// Both selections are canonical, i.e. sorted and non-overlapping. The dataset
```
mgeplf, line 165: I wonder if it's worth making `CanonicalSelection` a type, so there is no chance a non-canoical version is supplied?

include/bbp/sonata/population.h
```diff
@@ -194,7 +196,12 @@ template <typename Population>
 class SONATA_API PopulationStorage
 {
   public:
-    PopulationStorage(const std::string& h5FilePath, const std::string& csvFilePath = "");
+    PopulationStorage(const std::string& h5FilePath);
```
mgeplf, line 199: would having defaults for `csvFilePath` and `hdf5_reader` make this simpler?  I think it would simplify the `binds` as well, no?  I'm probably missing something.

src/read_bulk.hpp
mgeplf: let's get that merged, and then clean this one up

mgeplf (commented):
LGTM, let's get #319 merged, and then hopefully this

include/bbp/sonata/hdf5_reader.h
```diff
@@ -0,0 +1,197 @@
...
+///        hdf5_reader.readSelection(dset, selection);
+///      } else {
+///        hdf5_reader.readSelection(dset, {});
+///      }
+///    }
+///
+class Hdf5Reader
+{
+  public:
+    // The issue here is that on a mac `size_t` is different from
```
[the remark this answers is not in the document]
mgeplf: here; for example:
https://github.com/BlueBrain/libsonata/blob/master/src/nodes.cpp#L168
`src/population.cpp` has something similar

include/bbp/sonata/hdf5_reader.h
1uc: The difficulty would be functions like:
```
foo(const Selection& selection) {
  if(is_canonical(selection)) {
    // Broken ownership:
    auto sel = CanonicalSelection(std::move(selection.ranges());

    // Feels needlessly expensive, especially if we cross canonical/non-canonical
    // borders a couple of times.
    auto sel = CanonicalSelection(selection.ranges());
    bar(broken_ownership, expensive);
  }
  else {
    // Easy: copy is happening anyway because of the sorting and merging.
    bar(sortAndMerge(selection));
  }
}
```
Meaning I don't think we can do it late easily. If we try to do it early we might get stuck on functions like:
```
??? generateSelection(...) {
  // Depending on the arguments the created selection happens to
  // be canonical, but we can't tell from the arguments. Hence, we don't know
  // the return type.
} 
```

Naturally, we can have a:
```
class CanonicalSelection : public Selection {}
```
and then do things like:
```
auto sel = Selection();
if(is_canonical(sel)) {
  foo(static_cast<const CanonicalSelection&>(sel));
} else {
  foo(sortAndMerge(sel));
}
```
One can do cheap checks rough checks:
```
bool canonical = dynamic_cast<CanonicalSelection const * const>(&sel) != nullptr;
```
allowing one to do expensive sanity check everywhere by short-circuiting with the above trick. Not sure we'd get style point though.

include/bbp/sonata/population.h
1uc: Do we allow:
```
PopulationStorage("foo.h5", hdf5_reader);
```

If yes, we'll need at least:
```
    PopulationStorage(const std::string& h5FilePath, const std::string& csvFilePath = "", const Hdf5Reader& = Hdf5Reader());
    PopulationStorage(const std::string& h5FilePath, const Hdf5Reader& hdf5_reader);
```
I don't see how it'll help with the bindings.

I think there's a tiny difference that if the default is in the signature, it updates when recompiling the code using libsonata. When putting it inside the body of the function, the default should update when simply loading to a different version of libsonata.

Honestly, it's just the version I prefer if there's more then one overload.

mgeplf: > Honestly, it's just the version I prefer if there's more then one overload.

Ok, sounds good.

include/bbp/sonata/hdf5_reader.h
mgeplf: Ok, was an idle thought; we can try and bolt it on some other time, and see the ramifications.  Not needed for this PR.  Thanks for the analysis.

include/bbp/sonata/hdf5_reader.h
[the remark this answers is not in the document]
1uc: done.

mgeplf (approved):

PR #312, opened 2023-11-14
No 'brew update'.

1uc:


1uc:
Other failed attempts in #313. I propose we avoid updating, also because the error seems quite generic and could easily also happen outside of CI. The issue is that `brew update` installs into a system wide location which may conflict with existing installations.

PR #314, opened 2023-11-14
Refactor edge_index::resolve.

1uc:
The point is to split reading of the dataset into a separate function, and then
make `resolve` safe for collective IO (assuming the newly introduced function
is).

The overload for reading a single `nodeID` is removed as it's unused now.

### DIFF
--- a/src/edge_index.cpp
+++ b/src/edge_index.cpp
@@ -77,7 +77,6 @@ Selection resolve(const HighFive::Group& indexGroup, const std::vector<NodeID>& 
         return range[0] >= range[1];
     });
 
-    // TODO check that the spec allows us to optimize this.
     primaryRange = bulk_read::sortAndMerge(primaryRange);
 
     auto secondaryRange = detail::readCanonicalSelection<std::array<uint64_t, 2>>(
--- a/src/edge_index.cpp
+++ b/src/edge_index.cpp
@@ -18,6 +18,7 @@
 #include <vector>
 
 #include "read_bulk.hpp"
+#include "read_canonical_selection.hpp"
 
 namespace bbp {
 namespace sonata {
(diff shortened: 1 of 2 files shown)


1uc:
This is a preparatory PR for #307

src/edge_index.cpp
```diff
@@ -54,78 +55,33 @@ const HighFive::Group targetIndex(const HighFive::Group& h5Root) {
...
-        Selection::fromValues(nodeIDs),
-        min_gap_size,
-        max_aggregated_block_size);
-
-    // Sort and eliminate empty ranges.
+    auto node2ranges_dset = indexGroup.getDataSet(NODE_ID_TO_RANGES_DSET);
+    auto node_dim = node2ranges_dset.getSpace().getDimensions()[0];
+    auto sortedNodeIds = nodeIDs;
+    bulk_read::detail::erase_if(sortedNodeIds, [node_dim](auto id) {
+        // Filter out `nodeIDs[i] >= dims`; because SYN2 used to return an
```
matz-e, line 63: Where does SYN2 suddenly come from?

src/edge_index.cpp
```diff
@@ -54,78 +55,33 @@ const HighFive::Group targetIndex(const HighFive::Group& h5Root) {
     return h5Root.getGroup(TARGET_INDEX_GROUP);
 }
 
-Selection resolve(const HighFive::Group& indexGroup, const NodeID nodeID) {
-    if (nodeID >= indexGroup.getDataSet(NODE_ID_TO_RANGES_DSET).getSpace().getDimensions()[0]) {
-        // Returning empty set for out-of-range node IDs, to be aligned with SYN2 reader
-        // implementation
-        // TODO: throw a SonataError instead
-        return Selection({});
-    }
```
1uc, line 63: Here. @matz-e

matz-e: Thanks! We don't deal in SYN2 any longer, maybe we should invoke that TODO (@mgeplf ?)

src/edge_index.cpp
matz-e: Wait, we're just throwing these away now?

1uc: No, we're preserving the previous behaviour. An out-of-range Node ID results in an empty range. Multiple Node IDs were implemented by looping over all Node IDs, loading one Node ID at a time. The effect is that it ignored any out-of-range Node IDs.

Hence, if we perform the operation in bulk we simply filter out all out-of-range Node IDs. An load the valid ones. If there only out-of-range Node IDs, then it'll return an empty Selection.

1uc: See, https://github.com/BlueBrain/libsonata/pull/314#discussion_r1402326533

src/edge_index.cpp
mgeplf: I *think* (but I wouldn't my life on it) this is working around a somewhat ambiguous index case: whether or not the an index range points at the last element, or one past it.  I think keeping the behavior, and not throwing an exeption is fine.

src/edge_index.cpp
```diff
@@ -54,78 +55,33 @@ const HighFive::Group targetIndex(const HighFive::Group& h5Root) {
...
+    sortedNodeIds.erase(std::unique(sortedNodeIds.begin(), sortedNodeIds.end()),
+                        sortedNodeIds.end());
+
+    auto nodeSelection = Selection::fromValues(sortedNodeIds);
+    auto primaryRange = detail::readCanonicalSelection<std::array<uint64_t, 2>>(
+        node2ranges_dset, nodeSelection.ranges(), RawIndex{{0, 2}});
+
+    bulk_read::detail::erase_if(primaryRange, [](const auto& range) {
+        // Filter out any invalid ranges `start >= end`.
+        return range[0] >= range[1];
```
mgeplf, line 77: how hard would it be to have a test for this? I see that we're kinda lacking the `tests/test_edges.cpp` front, but maybe it's a chance to start adding some?

src/read_canonical_selection.hpp
```diff
@@ -0,0 +1,38 @@
...
+                                      const std::vector<XRange>& xranges,
+                                      const std::vector<YRange>& yranges) {
+    if (yranges.size() != 1) {
+        throw SonataError("Only yranges.size() == 1 has been implemented.");
+    }
+
+    size_t j_begin = std::get<0>(yranges[0]);
+    size_t j_end = std::get<1>(yranges[0]);
+
+    constexpr size_t min_gap_size = SONATA_PAGESIZE / (2 * sizeof(uint64_t));
```
mgeplf, line 20: will `SONATA_PAGESIZE` show up in other files? I'd prefer to move it to this one if it's only used here.

mgeplf (commented):
Not complains on my side, looks good.

src/read_canonical_selection.hpp
1uc: If continue with the set of changes, then this function will move into a virtual method. Which means the constant can move into the class too, and be a runtime constant, not a compile time constant.

src/edge_index.cpp
```diff
@@ -54,78 +55,33 @@ const HighFive::Group targetIndex(const HighFive::Group& h5Root) {
...
+    });
+
+    // TODO check that the spec allows us to optimize this.
```
1uc, line 80: Obviously doesn't.

src/edge_index.cpp
1uc: Isn't this already tested here:
https://github.com/BlueBrain/libsonata/blob/bfd817f3c5a939bed92e826f01993b6305bad263/tests/test_edges.cpp#L49
https://github.com/BlueBrain/libsonata/blob/bfd817f3c5a939bed92e826f01993b6305bad263/tests/test_edges.cpp#L55

The indices are:
```
$ h5dump -d edges/edges-AB/indices/target_to_source/node_id_to_ranges tests/data/edges1.h5 
HDF5 "tests/data/edges1.h5" {
DATASET "edges/edges-AB/indices/target_to_source/node_id_to_ranges" {
   DATATYPE  H5T_STD_U64LE
   DATASPACE  SIMPLE { ( 4, 2 ) / ( 4, 2 ) }
   DATA {
   (0,0): 0, 1,
   (1,0): 1, 3,
   (2,0): 3, 5,
   (3,0): 5, 5
   }
}
}

$ h5dump -d edges/edges-AB/indices/source_to_target/node_id_to_ranges tests/data/edges1.h5
HDF5 "tests/data/edges1.h5" {
DATASET "edges/edges-AB/indices/source_to_target/node_id_to_ranges" {
   DATATYPE  H5T_STD_U64LE
   DATASPACE  SIMPLE { ( 4, 2 ) / ( 4, 2 ) }
   DATA {
   (0,0): 0, 0,
   (1,0): 0, 1,
   (2,0): 1, 2,
   (3,0): 2, 3
   }
}
}
```
Meaning `0` and `3` are empty ranges.

src/read_canonical_selection.hpp
mgeplf: Yup, sounds good.

src/edge_index.cpp
mgeplf: I suppose; since the functions can be called w/o an hdf5 file, it would probably make it easier to make changes in the future if they had explicit tests, but we can skip that for now.

mgeplf (approved):

PR #315, opened 2023-11-14
Refactor _readSelection.

1uc:
This commit refactors `_readSelection` in such a manner that:
  1. Canonical selections don't require post-read shuffling.
  2. The reading of the dataset is moved to a separate function.

As a consequence, there's no need for an optimization for `std::string`, since
those only matters when reading from the `"@library"`, which should only happen
indirectly. Hence, we free to ensure that those are always canonical.

### DIFF
--- a/src/population.hpp
+++ b/src/population.hpp
@@ -20,6 +20,8 @@
 
 #include <fmt/format.h>
 
+#include "read_bulk.hpp"
+#include "read_canonical_selection.hpp"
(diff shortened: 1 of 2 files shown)


1uc:
This is a preparatory PR for https://github.com/BlueBrain/libsonata/pull/307

mgeplf (approved):
LGTM

PR #319, opened 2023-11-23
Use `using Range = std::array<Value, 2>`.

1uc:
By using an `std::array` we can removed the differences between `RawIndex` and `Selection::Ranges`. It also allows us to read `Selection::Ranges` from disk directly via the HighFive API.

### DIFF
--- a/include/bbp/sonata/selection.h
+++ b/include/bbp/sonata/selection.h
@@ -14,7 +14,7 @@ class SONATA_API Selection
   public:
     using Value = uint64_t;
     using Values = std::vector<Value>;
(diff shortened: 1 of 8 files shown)


1uc:
This PR contains two variants:
* The first one allows pybind11 to pick the Python type which changes `type(selection.ranges)` from `List[Tuple]` to `List[List]` . 
* The second one creates the copy manually and creates a `List[Tuple]` as before.

Note, that the default STL conversion implies a copy. Hence, this is likely the same amount of copies.

matz-e (dismissed):

mgeplf (approved):

PR #326, opened 2023-12-06
Support collective reader.

1uc:
This allows:
```
from libsonata import make_collective_reader
```
Therefore any code that wants to use collective I/O can easily use the collective reader and it will automatically work even if their users can't/didn't install `libsonata_mpi`.

Unfortuntately, we can't mix `libsonata_mpi` and `libsonata` built from wheels, because the one needs pHDF5 and the other doesn't provide those. Hence, in the limited sites where collective IO is required, i.e. clusters with parallel filesystems, one needs to build both from source against the same version of HDF5.

### DIFF
--- a/python/libsonata/__init__.py
+++ b/python/libsonata/__init__.py
@@ -47,3 +47,22 @@ __all__ = [
     "version",
     "Hdf5Reader",
 ]
+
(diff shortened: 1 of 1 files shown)


1uc:
Been moved to BlueBrain/libsonata-mpi.

mgeplf (approved):

PR #330, opened 2023-12-12
Deduplicate CI.

1uc:


mgeplf (approved):
Let's give it a try; I swear I've tried various incarnations of this, but I don't remember - I hate the waste of doing it twice.

PR #333, opened 2023-12-12
Check more variants of CMake invocations.

1uc:
The failure of:
```
cmake --build build --target install
```
from a freshly configured `build` caused these additions.


1uc:
I suspect this will start working if #336 gets merged. This could also be closed if it's deemed too pedantic.

mgeplf (approved):
I like correctness, and testing that this works makes sense to me.

PR #335, opened 2023-12-12
Allow `sonata::sonata_{shared,static}`.

1uc:
When using libsonata as a submodule the targets were called `sonata_{shared,static}`. When using it via `find_package` they were called `sonata::sonata_{shared,static}`.

This commit introduces aliases for `sonata::sonata_{shared,static}`.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -138,6 +138,8 @@ foreach(TARGET sonata_shared sonata_static)
             PRIVATE gcov
         )
     endif()
+
+    add_library(sonata::${TARGET} ALIAS ${TARGET})
 endforeach(TARGET)
 
 set_target_properties(sonata_shared


1uc:
The issue is CMake code consuming `libsonata`. Those might have something like:
```
project(Foo)

if(FOO_FROM_SUBMODULES)
  add_subdirectory(extlib/libsonata)
else()
  find_package(sonata)
endif()

add_executable(foo foo.cpp)

# wrong when `FOO_FROM_SUBMODULES`
target_link_libraries(foo PRIVATE sonata::sonata_shared)

# wrong when `NOT FOO_FROM_SUBMODULES`
target_link_libraries(foo PRIVATE sonata_shared)
```

mgeplf (approved):

PR #338, opened 2023-12-14
Make sure defaults for condition blocks and others are applied

mgeplf:


joni-herttuainen (approved):
It seems to me that the defaults use the values in the spec. Also: I did not realize that `Simulation.network` defaults to `circuit_config.json`. Need to change that in the validator.

jorblancoa (approved):

PR #359, opened 2024-07-01
SimulationConfig: modifications as a list to preserve the order

WeinaJi:
SPEC: https://github.com/BlueBrain/sonata-extension/pull/55


src/config.cpp
```diff
@@ -629,16 +634,18 @@ void parseConditionsModifications(const nlohmann::json& it,
         case SimulationConfig::ModificationBase::ModificationType::TTX: {
             SimulationConfig::ModificationTTX result;
             result.type = type;
+            parseMandatory(valueIt, "name", debugStr, result.name);
             parseMandatory(valueIt, "node_set", debugStr, result.nodeSet);
-            buf[mIt.key()] = result;
+            buf.push_back(std::move(result));
```
jorblancoa, line 639: maybe use emplace_back instead?

src/config.cpp
```diff
@@ -629,16 +634,18 @@ void parseConditionsModifications(const nlohmann::json& it,
...
             break;
         }
         case SimulationConfig::ModificationBase::ModificationType::ConfigureAllSections: {
             SimulationConfig::ModificationConfigureAllSections result;
             result.type = type;
+            parseMandatory(valueIt, "name", debugStr, result.name);
             parseMandatory(valueIt, "node_set", debugStr, result.nodeSet);
             parseMandatory(valueIt, "section_configure", debugStr, result.sectionConfigure);
-            buf[mIt.key()] = result;
+            buf.push_back(std::move(result));
```
jorblancoa, line 648: same than avobe

src/config.cpp
WeinaJi: As we use `std::move` here, it avoids copy, right? Is `emplace_back` still better?

jorblancoa: you are right it is equivalent in this case...ignore please!

jorblancoa (approved):

PR #360, opened 2024-07-14
include fmt/ranges.h for using fmt::join()

tchaikov:
fmt::join() was moved into fmt/ranges.h in fmt 11, so let's include this header file also.


tchaikov:
hi @WeinaJi @mgeplf could you help review this change at your convenience? i see you've been contributing this project, so wondered if you could help on this change.

mgeplf:
Thanks for the change.

PR #365, opened 2024-07-29
Update HighFive to 2.10.0.

1uc:
Versions prior to `2.10.0` use HDF5 in a way that causes quadratic runtime when constructing a selection from many irregular hyperslabs.


mgeplf (approved):

mgeplf:
wow, thanks for that.

PR #369, opened 2024-09-04
Update upload/download action to `v4`.

1uc:


.github/workflows/publish-sdist-wheels.yml
```diff
@@ -139,10 +139,11 @@ jobs:
 
     steps:
       - name: Download artifacts produced during the build_wheels and build_sdist jobs
-        uses: actions/download-artifact@v3
+        uses: actions/download-artifact@v4
         with:
-          name: dist
+          name: dist-*
           path: dist/
+          merge-multiple: true
```
mgeplf, line 146: First of all, thanks for looking at this drudgery!

The one thing that isn't tested unless we do a release is the upload to pypi.
I *suspect* that `merge-multiple` covers that base; but I'm not 100% sure - I assume that's your expectation as well?

I'm fine with waiting until we actually need to do a release for that to be "tested".

mgeplf (approved):

PR #374, opened 2024-10-17
Use aggregated reading for 1D selections.

1uc:
Aggregating scattered reads was implemented in

  * 836646574302c2438a8455595d996c5be52a84a8 adds bulkRead,
  * a7d1453fadce88201d009ea3c843f562491238bf for 2D selections.

However, it was only used for edge indexes. We measured the performance on an HDD and SSD for random access patterns with different streak lengths. We found that for all cases, including streak length of 1, using bulkRead was preferrable over both `0.1.24` (using `HighFive::ElementSet`) and `0.1.29` (using `HighFive::HyperSlab`).

We investigated if aggregating multiple pages was beneficial and found that it was not.


mgeplf:
Impressive.

mgeplf (approved):

